PR #1, opened 2023-10-16
Init project

GromNaN:
Import code from https://github.com/mongodb/mongo-php-library/pull/1180


.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,94 @@
...
+
+    php -i | grep -E 'mongodb|libmongoc|libbson'
+
+In some cases, it may be helpful to provide all phpinfo() output (e.g. `php -i`)
+and/or list all installed extensions (e.g. `php -m`).
+
+Include version information for the library. This may be
+collected by running the following from your project's root directory:
+
+    composer show mongodb/mongodb
```
jmikola: Should this be left as-is, or would you want to add a second line to dump the builder version as well?

.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,94 @@
...
+and/or `var_dump()` output if that helps explain the error.
+
+If the observed behavior is an unexpected exception, please include its full
+message and any relevant backtrace information. If you encountered a
+segmentation fault, please include a GDB backtrace[^4].
+
+[^4]: https://bugs.php.net/bugs-generating-backtrace.php
+-->
+
+### Debug Log
```
jmikola: Not sure if this section applies to the builder, since we're now two levels above the extension.

.github/workflows/coding-standards.yml
```diff
@@ -0,0 +1,63 @@
...
+  push:
+    branches:
+      - "[0-9]+.[0-9]+"
+      - "feature/*"
+    paths-ignore:
+      - "docs/**"
+
+env:
+  PHP_VERSION: "8.2"
+  DRIVER_VERSION: "mongodb/mongo-php-driver@master"
```
jmikola, line 19: Does this actually require 1.17-dev, or was this just copied from PHPLIB? I would think we can rely on the stable release here.

.github/workflows/static-analysis.yml
```diff
@@ -0,0 +1,62 @@
...
+  push:
+    branches:
+      - "[0-9]+.[0-9]+"
+      - "feature/*"
+    paths-ignore:
+      - "docs/**"
+
+env:
+  PHP_VERSION: "8.2"
+  DRIVER_VERSION: "mongodb/mongo-php-driver@master"
```
jmikola, line 19: Same question as in `coding-standards.yml`. Can this just use the most recent stable release?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,112 @@
...
+        os:
+          - "ubuntu-20.04"
+        php-version:
+          - "8.1"
+          - "8.2"
+          - "8.3"
+        mongodb-version:
+          - "4.4"
+        driver-version:
+          - "mongodb/mongo-php-driver@master"
```
jmikola, line 34: Should this (and `driver-version` instances below) use the stable release?

README.md
```diff
@@ -1 +1,21 @@
...
+$ composer require mongodb/builder
+```
+
+## Usage
+
+
+## Development
+
+Development is tracked in the
+[PHPLIB](https://jira.mongodb.org/projects/PHPLIB/summary) project in MongoDB's
```
jmikola, line 19: I assume PHPLIB is still preferable to PHPORM here.

composer.json
```diff
@@ -10,7 +10,9 @@
         { "name": "Jérôme Tamarelle", "email": "jerome.tamarelle@mongodb.com" }
     ],
     "require": {
-        "php": "^8.1"
+        "php": "^8.1",
+        "ext-mongodb": "^1.17.0",
```
jmikola, line 14: Ah, it just occurred to me that extension version 1.17-dev is required because we're depending on the dev branch of PHPLIB. This likely answers my earlier questions about the CI configs, but I'll defer to you to resolve those (no comment needed).

tests/Builder/PipelineTest.php
```diff
@@ -0,0 +1,54 @@
...
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('Named arguments are not supported for pipelines');
+
+        new Pipeline(
+            $this->createMock(StageInterface::class),
+            foo: $this->createMock(StageInterface::class),
+        );
+    }
+
+    public function testRejectEmptyPipeline(): void
```
jmikola: If we ultimately expect to support passing a Pipeline object in places where PHP arrays are supported (e.g. `watch()` and `aggregate()` methods in the library), then I think an empty pipeline should be valid.

composer.json
```diff
@@ -10,7 +10,9 @@
         { "name": "Jérôme Tamarelle", "email": "jerome.tamarelle@mongodb.com" }
     ],
     "require": {
-        "php": "^8.1"
+        "php": "^8.1",
```
jmikola: Why does the generator use `>=8.1` instead of `^8.1`? Should these be consistent?

tests/Builder/BuilderEncoderTest.php
```diff
@@ -0,0 +1,306 @@
...
+        $this->assertSamePipeline($expected, $pipeline);
+    }
+
+    /** @see https://www.mongodb.com/docs/manual/reference/operator/aggregation/match/#perform-a-count */
+    public function testPerformCount(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::match(
+                Query::or(
+                    Query::query(score: [Query::gt(70), Query::lt(90)]),
```
jmikola, line 62: I'm not sure if this file is the right place for it, but should there be an error test for providing an array with two identical operators? For example:

```php
Query::query(score: [Query::gt(70), Query::gt(50)]),
```

tests/Builder/BuilderEncoderTest.php
```diff
@@ -0,0 +1,306 @@
...
+            ],
+        ];
+
+        $this->assertSamePipeline($expected, $pipeline);
+    }
+
+    /**
+     * @see https://www.mongodb.com/docs/manual/reference/operator/aggregation/filter/#examples
+     *
+     * @dataProvider provideExpressionFilterLimit
```
jmikola, line 95: If you don't anticipate the data provider growing this could potentially use [`@testWith`](https://docs.phpunit.de/en/10.4/annotations.html#testwith). Feel free to resolve w/o action.

tests/Builder/BuilderEncoderTest.php
```diff
@@ -0,0 +1,306 @@
...
+        ];
+
+        yield 'int limit' => [
+            [1],
+            ['limit' => 1],
+        ];
+    }
+
+    /** @see https://www.mongodb.com/docs/manual/reference/operator/aggregation/slice/#example */
+    public function testSlice(): void
```
jmikola, line 141: Would it make more sense to split up these tests by operator? I presume you might also use namespaces to group the tests.

tests/Builder/BuilderEncoderTest.php
```diff
@@ -0,0 +1,306 @@
...
+        ];
+
+        $this->assertSamePipeline($expected, $pipeline);
+    }
+
+    public function testUnionWith(): void
+    {
+        $unionWith = Stage::unionWith(...);
+        $match = Stage::match(...);
+        $project = Stage::project(...);
```
jmikola: What is the purpose of using [first class callable syntax](https://www.php.net/manual/en/functions.first_class_callable_syntax.php) here? These callables are only referenced once below.

tests/Builder/BuilderEncoderTest.php
```diff
@@ -0,0 +1,306 @@
...
+    /** @see https://www.mongodb.com/docs/manual/reference/operator/aggregation/redact/ */
+    public function testRedactStage(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::match(status: 'A'),
+            Stage::redact(
+                Expression::cond(
+                    if: Expression::eq(Expression::fieldPath('level'), 5),
+                    then: '$$PRUNE',
+                    else: '$$DESCEND',
```
jmikola: This may be an opportunity to test/demonstrate Variable.

src/Builder/Type/QueryObject.php
```diff
@@ -0,0 +1,87 @@
...
+        Serializable|array|stdClass $queries,
+    ) {
+        $seenQueryOperators = [];
+        foreach ($queries as $fieldPath => $query) {
+            if ($query instanceof QueryInterface) {
+                if (! is_numeric($fieldPath)) {
+                    throw new InvalidArgumentException(sprintf('Query operator "%s" cannot be used with a field path. Got "%s".', $query::NAME, $fieldPath));
+                }
+
+                if (! $query instanceof self) {
```
jmikola: Is it reasonable to assume that if `$query` is a QueryInterface and not a QueryObject, it will always be another implementation that defines a `NAME` constant? That seems fragile.

src/Builder/Type/QueryObject.php
```diff
@@ -0,0 +1,87 @@
...
+            // Convert list of filters into $and
+            if (self::isListOfFilters($query)) {
+                if (count($query) === 1) {
+                    $query = $query[0];
+                } else {
+                    $query = new CombinedFieldQuery($query);
+                }
+            }
+
+            // Numbers are valid field paths, nothing to validate.
```
jmikola, line 65: What's the significance of this comment?

Earlier, you raise an exception if `$fieldPath` is _not_ numeric within the `$query instanceof QueryInterface` branch.

Isn't it possible that we get to this point and `$fieldPath` _is_ numeric?

src/Builder/Type/OutputWindow.php
```diff
@@ -0,0 +1,70 @@
...
+ * Specifies the window boundaries and parameters. Window boundaries are inclusive.
+ * Default is an unbounded window, which includes all documents in the partition.
+ *
+ * @see https://www.mongodb.com/docs/manual/reference/operator/aggregation/setWindowFields/
+ */
+class OutputWindow implements WindowInterface
+{
+    public const ENCODE = Encode::Object;
+
+    /** @param Document|Serializable|WindowInterface|array|stdClass $operator Function used to initialize the state. The init function receives its arguments from the initArgs array expression. You can specify the function definition as either BSON type Code or String. */
```
jmikola, line 25: I don't think `@param` applies to a class property. Isn't `@var` typically used?

In any event, repeating the union type here seems redundant.

src/Builder/Type/OutputWindow.php
```diff
@@ -0,0 +1,70 @@
...
+     * @param Document|Serializable|WindowInterface|array|stdClass $operator  Window operator to use in the $setWindowFields stage.
+     * @param Optional|array{string|int,string|int}                $documents A window where the lower and upper boundaries are specified relative to the position of the current document read from the collection.
+     * @param Optional|array{string|numeric,string|numeric}        $range     Arguments passed to the init function.
+     * @param Optional|non-empty-string                            $unit      Specifies the units for time range window boundaries. If omitted, default numeric range window boundaries are used.
+     */
+    public function __construct(
+        Document|Serializable|WindowInterface|stdClass|array $operator,
+        Optional|array $documents = Optional::Undefined,
+        Optional|array $range = Optional::Undefined,
+        Optional|string $unit = Optional::Undefined,
```
jmikola, line 40: Is `Optional::Undefined` used everywhere for optional values? I thought it might only be required to avoid ambiguity when `null` is a valid value.

src/Builder/Type/OutputWindow.php
```diff
@@ -0,0 +1,70 @@
...
+    ) {
+        $this->operator = $operator;
+
+        $window = null;
+        if ($documents !== Optional::Undefined) {
+            if (! array_is_list($documents) || ! count($documents) === 2) {
+                throw new InvalidArgumentException('Expected $documents argument to be a list of 2 string or int.');
```
jmikola, line 47:
```suggestion
                throw new InvalidArgumentException('Expected $documents argument to be a list of 2 string or int');
```

No need for periods in single-sentence exception messages.

src/Builder/Type/OutputWindow.php
```diff
@@ -0,0 +1,70 @@
...
+            }
+
+            $window ??= new stdClass();
+            $window->documents = $documents;
+        }
+
+        if ($range !== Optional::Undefined) {
+            if (! array_is_list($range) || ! count($range) === 2) {
+                throw new InvalidArgumentException('Expected $range argument to be a list of 2 string or numeric.');
```
jmikola, line 56: Where is `string|numeric` being checked? Likewise for `string|int` above when checking `$documents`.

src/Builder/Type/OutputWindow.php
```diff
@@ -0,0 +1,70 @@
...
+     */
+    public function __construct(
+        Document|Serializable|WindowInterface|stdClass|array $operator,
+        Optional|array $documents = Optional::Undefined,
+        Optional|array $range = Optional::Undefined,
+        Optional|string $unit = Optional::Undefined,
+    ) {
+        $this->operator = $operator;
+
+        $window = null;
```
jmikola, line 44: I don't think this assignment is actually necessary. `??` and `??=` handle undefined vars.

src/Builder/Type/Optional.php
```diff
@@ -0,0 +1,17 @@
...
+
+declare(strict_types=1);
+
+namespace MongoDB\Builder\Type;
+
+/**
+ * Allow to omit optional arguments in operator builders.
+ * NULL cannot be used to mark an argument as not set, because it is a valid value for some operators.
+ *
+ * @internal
```
jmikola, line 11: Just to confirm: there should never be a case where users try to pass `Optional::Undefined`?

src/Builder/Type/Encode.php
```diff
@@ -0,0 +1,19 @@
...
+namespace MongoDB\Builder\Type;
+
+/**
+ * Defines how to encode a stage or an operator into BSON.
+ *
+ * @see BuilderEncoder
+ */
+enum Encode
+{
+    // @todo add comments (see schema.json)
```
jmikola, line 14: Reminder to address this TODO.

src/Builder/Type/Encode.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Builder\Type;
+
+/**
+ * Defines how to encode a stage or an operator into BSON.
+ *
+ * @see BuilderEncoder
```
jmikola, line 10: This exists in a different namespace: MongoDB\Builder. Do you need a `use` statement?

src/Builder/Type/CombinedFieldQuery.php
```diff
@@ -0,0 +1,30 @@
...
+ * List of filters that apply to the same field path.
+ */
+class CombinedFieldQuery implements FieldQueryInterface
+{
+    /** @param list<FieldQueryInterface|Serializable|array|stdClass> $filters */
+    public function __construct(public array $filters)
+    {
+        foreach ($filters as $filter) {
+            if (! $filter instanceof FieldQueryInterface && ! $filter instanceof Serializable && ! is_array($filter) && ! $filter instanceof stdClass) {
+                throw new InvalidArgumentException(sprintf('Expected filters to be a list of %s, %s, array or stdClass, %s given.', FieldQueryInterface::class, Document::class, get_debug_type($filter)));
```
jmikola: Do you need to assert that `$filters` is a list, or at least normalize it with `array_values()`? As-is, the user could pass an associative array.

src/Builder/Type/CombinedFieldQuery.php
```diff
@@ -0,0 +1,30 @@
...
+use function is_array;
+use function sprintf;
+
+/**
+ * List of filters that apply to the same field path.
+ */
+class CombinedFieldQuery implements FieldQueryInterface
+{
+    /** @param list<FieldQueryInterface|Serializable|array|stdClass> $filters */
+    public function __construct(public array $filters)
```
jmikola: Is this intended to be a public property? Should it be `readonly` if so?

src/Builder/Stage.php
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * Filters the document stream to allow only matching documents to pass unmodified into the next pipeline stage. $match uses standard MongoDB queries. For each input document, outputs either one document (a match) or zero documents (no match).
+     *
+     * @see https://www.mongodb.com/docs/manual/reference/operator/aggregation/match/
+     *
+     * @param Document|FieldQueryInterface|QueryInterface|Serializable|array|stdClass $queries
+     */
+    public static function match(FieldQueryInterface|QueryInterface|Serializable|array|bool|float|int|stdClass|string|null ...$queries): MatchStage
+    {
+        // Override the generated method to allow variadic arguments
```
jmikola, line 29: Is there a reason the generated method doesn't accept var args on its own?

src/Builder/Stage.php
```diff
@@ -0,0 +1,32 @@
...
+    use Stage\FactoryTrait {
+        match as private generatedMatch;
+    }
+
+    /**
+     * Filters the document stream to allow only matching documents to pass unmodified into the next pipeline stage. $match uses standard MongoDB queries. For each input document, outputs either one document (a match) or zero documents (no match).
+     *
+     * @see https://www.mongodb.com/docs/manual/reference/operator/aggregation/match/
+     *
+     * @param Document|FieldQueryInterface|QueryInterface|Serializable|array|stdClass $queries
```
jmikola: Why does this `@param` type union omit `null`? Does it serve any purpose if it just duplicates the actual type info below?

.github/ISSUE_TEMPLATE/bug-report.md
GromNaN: I'll remove this file and disable issues on this project.

.github/workflows/coding-standards.yml
GromNaN: We use the codec feature from unreleased 1.17

README.md
GromNaN: Yes, this package will be merged into PHPLIB later.

composer.json
GromNaN: Fixed.

tests/Builder/PipelineTest.php
GromNaN: You're right, an empty pipeline is perfectly valid. Updated.

tests/Builder/BuilderEncoderTest.php
GromNaN: I don't like `@testWith`. It doesn't allow to run a single test case in PHPStorm.
This tests will be reworked in [PHPLIB-1271](https://jira.mongodb.org/browse/PHPLIB-1271)

composer.json
OskarStark: I always go with `>=8.1` in open source projects, as I don't know the compatibility of new versions yet and I don't want to block users. As this is a repo by a big company this problem should not occur 😄

generator/composer.json
```diff
@@ -0,0 +1,34 @@
...
+    },
+    "require": {
+        "php": ">=8.1",
+        "ext-mongodb": "^1.16.0",
+        "mongodb/mongodb": "^1.17.0@dev",
+        "mongodb/builder": "@dev",
+        "nette/php-generator": "^4",
+        "symfony/console": "^6.3",
+        "symfony/finder": "^6.3",
+        "symfony/yaml": "^6.3"
```
OskarStark: You can allow Symfony 7 here

src/Builder/Accumulator/FactoryTrait.php
```diff
@@ -0,0 +1,512 @@
...
+use stdClass;
+
+/**
+ * @internal
+ */
+trait FactoryTrait
+{
+    /**
+     * Defines a custom accumulator function.
+     * New in version 4.4.
```
OskarStark: Which package version is meant here?

src/Builder/Accumulator/FactoryTrait.php
```diff
@@ -0,0 +1,512 @@
...
+    public static function addToSet(
+        Type|ExpressionInterface|stdClass|array|bool|float|int|null|string $expression,
+    ): AddToSetAccumulator
+    {
+        return new AddToSetAccumulator($expression);
+    }
+
+    /**
+     * Returns an average of numerical values. Ignores non-numeric values.
+     * Changed in version 5.0: Available in the $setWindowFields stage.
```
OskarStark: Same here, not sure those comments should be removed in the new package

src/Builder/Accumulator/FactoryTrait.php
GromNaN: It's MongoDB version. It's confusing, I admit.
Better like this?
```suggestion
     * New in MongoDB 4.4.
```

OskarStark: Yes

src/Builder/Stage.php
GromNaN: It's something I can work on later. Query can be passed to operators in all forms:
- As single param: `$elemMatch`, `$match`. Variadic is great for them.
- As a list of query: `$and`, `$or`, `$nor`, Variadic can't be used. But a variadic list of `query(/* ... */)`
- With other params: `$geoNear`, `$graphLookup`. Variadic can't be used.
- `$not` accept a single fieldQuery object without field name. So not variadic.

And I need to wrap the query objects in `QueryObject` and `CombinedFieldQuery`

.github/ISSUE_TEMPLATE/bug-report.md
GromNaN: Removed this file, we disabled issues on this project.

composer.json
GromNaN: Fixed.

tests/Builder/BuilderEncoderTest.php
GromNaN: That's what `CombinedFieldQuery` does.

src/Builder/Stage.php
GromNaN: Fixed.

src/Builder/Type/CombinedFieldQuery.php
GromNaN: Yes. Fixed.

tests/Builder/BuilderEncoderTest.php
GromNaN: Part of [PHPLIB-1271](https://jira.mongodb.org/browse/PHPLIB-1271)

tests/Builder/BuilderEncoderTest.php
GromNaN: Leftover after experimentation. https://dev.to/gromnan/how-to-override-phps-naming-constraints-to-model-mongodb-operators-3p30#closures-in-variables

tests/Builder/BuilderEncoderTest.php
GromNaN: Fixed. I'll add factories for common variables. [PHPLIB-1282](https://jira.mongodb.org/browse/PHPLIB-1282)

tests/Builder/BuilderEncoderTest.php
jmikola: OK, so I think my ask here is test coverage for that class. I assume we'll get that down the line, so feel free to resolve this thread for now.

generator/config/expressions.php
```diff
@@ -0,0 +1,163 @@
...
+    'string' => ['string'],
+    'object' => ['array', stdClass::class, BSON\Document::class, BSON\Serializable::class],
+    'array' => ['array', BSONArray::class, BSON\PackedArray::class],
+    'binData' => ['string', BSON\Binary::class],
+    'objectId' => [BSON\ObjectId::class],
+    'bool' => ['bool'],
+    'date' => [BSON\UTCDateTime::class],
+    'null' => ['null'],
+    'regex' => [BSON\Regex::class],
+    'javascript' => ['string', BSON\Javascript::class],
```
jmikola, line 32: I wasn't aware that aggregation framework accepted a string alongside a Javascript object, but I assume this pertains to [$function](https://www.mongodb.com/docs/manual/reference/operator/aggregation/function/#mongodb-expression-exp.-function).

src/Builder/Query/WhereOperator.php
```diff
@@ -0,0 +1,39 @@
...
+{
+    public const ENCODE = Encode::Single;
+
+    /** @var non-empty-string $function */
+    public readonly string $function;
+
+    /**
+     * @param non-empty-string $function
+     */
+    public function __construct(string $function)
```
jmikola, line 30: [$where](https://www.mongodb.com/docs/manual/reference/operator/query/where/) accepts both a string and a Javascript object.

tests/Builder/BuilderEncoderTest.php
GromNaN: I added `CombinedFieldQueryTest`.

src/Builder/Projection/ElemMatchOperator.php
```diff
@@ -0,0 +1,51 @@
...
+    /** @var Document|QueryInterface|Serializable|array|stdClass $query */
+    public readonly Document|Serializable|QueryInterface|stdClass|array $query;
+
+    /**
+     * @param Document|QueryInterface|Serializable|array|stdClass $query
+     */
+    public function __construct(Document|Serializable|QueryInterface|stdClass|array $query)
+    {
+        if (is_array($query) || is_object($query)) {
+            $query = QueryObject::create(...$query);
```
jmikola, line 41: Is it valid to use `...` on an arbitrary object? [Variable-length argument lists](https://www.php.net/manual/en/functions.arguments.php#functions.variable-arg-list) notes the `...` operator is only applicable to arrays and Traversable objects.

src/Builder/BuilderEncoder.php
```diff
@@ -0,0 +1,293 @@
...
+     * @see https://www.mongodb.com/docs/manual/reference/operator/aggregation/setWindowFields/
+     */
+    private function encodeOutputWindow(OutputWindow $outputWindow): stdClass
+    {
+        $result = $this->recursiveEncode($outputWindow->operator);
+
+        // Transform the result into an stdClass if a document is provided
+        if (! $outputWindow->operator instanceof WindowInterface && (is_array($result) || is_object($result))) {
+            if (! is_first_key_operator($result)) {
+                throw new LogicException(sprintf('Expected OutputWindow::$operator to be an operator. Got "%s"', array_key_first($result)));
```
jmikola, line 247: [`array_key_first()`](https://www.php.net/array_key_first) requires an array parameter. This would fail if `$result` is an object (possible by the condition above).

src/Builder/BuilderEncoder.php
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    /**
+     * Encode the value as an array of properties, in the order they are defined in the class.
+     */
+    private function encodeAsArray(OperatorInterface $value): stdClass
+    {
+        $result = [];
+        /** @var mixed $val */
+        foreach (get_object_vars($value) as $val) {
```
jmikola, line 130: Is this fragile in that it depends on the declaration order of class properties? I suppose if the classes are all generated from YAML this won't be a concern.

jmikola (approved):
A few comments that may highlight things to fix, but happy to get this merged and revisit in more detail once we start adding test coverage for the operators.

src/Builder/Type/QueryObject.php
GromNaN: I updated to check for `OperatorInterface`.

src/Builder/Type/OutputWindow.php
GromNaN: It's not easy to sort out in the codec afterwards. I prefer consistency.

src/Builder/Type/QueryObject.php
GromNaN: I think what I meant was that we don't need to check whether `$fieldPath` is an int or a string. It's the type of `$query` that will indicate whether it corresponds to a field name (when `FieldQueryInterface`)  or not (when `QueryInterface`).

src/Builder/Type/OutputWindow.php
GromNaN: Fixed.

src/Builder/Type/OutputWindow.php
GromNaN: Added checks.

src/Builder/Type/OutputWindow.php
GromNaN: Psalm want the variable to be declared.

src/Builder/Type/Optional.php
GromNaN: They should not pass the value to the function.

src/Builder/Type/CombinedFieldQuery.php
GromNaN: Assertion on `array_is_list` was added.

src/Builder/Query/WhereOperator.php
GromNaN: Type updated.

src/Builder/Projection/ElemMatchOperator.php
GromNaN: This will be reworked when adding tests to this operators.

src/Builder/BuilderEncoder.php
GromNaN: Yes, we use the order in which properties are declared in the object to generate the array. Otherwise, we have to generate a specific encoding class for each operator with all the necessary metadata.

PR #3, opened 2023-10-17
Validate duplicate operators are not used in CombinedFieldQuery

GromNaN:
In MongoDB syntax, it's not possible to use the same operator twice. The last value in the object replaces the previous one. 

```ts
{ name: { $eq: "Mars", $eq: "Jupiter" } }
// Same as
{ name: { $eq: "Jupiter" } }
```

This PR proposes to prohibit the use of the same operator 2 times. But we could also choose to implement the same logic as in JS.

### DIFF
--- a/src/Builder/Type/CombinedFieldQuery.php
+++ b/src/Builder/Type/CombinedFieldQuery.php
@@ -4,28 +4,84 @@ declare(strict_types=1);
 
 namespace MongoDB\Builder\Type;
 
-use MongoDB\BSON\Document;
 use MongoDB\BSON\Serializable;
 use MongoDB\Exception\InvalidArgumentException;
 use stdClass;
 
+use function array_is_list;
+use function array_key_exists;
+use function array_key_first;
+use function array_merge;
+use function array_reduce;
+use function count;
 use function get_debug_type;
+use function get_object_vars;
 use function is_array;
 use function sprintf;
+use function str_starts_with;
 
 /**
- * List of filters that apply to the same field path.
+ * List of field query that apply to the same field path.
  */
 class CombinedFieldQuery implements FieldQueryInterface
 {
-    public function __construct(
-        /** @var list<FieldQueryInterface|Serializable|array|stdClass> $fieldQueries */
-        public readonly array $fieldQueries,
-    ) {
-        foreach ($fieldQueries as $fieldQuery) {
-            if (! $fieldQuery instanceof FieldQueryInterface && ! $fieldQuery instanceof Serializable && ! is_array($fieldQuery) && ! $fieldQuery instanceof stdClass) {
-                throw new InvalidArgumentException(sprintf('Expected filters to be a list of %s, %s, array or stdClass, %s given.', FieldQueryInterface::class, Document::class, get_debug_type($fieldQuery)));
+    /** @var list<FieldQueryInterface|Serializable|array|stdClass> $fieldQueries */
+    public readonly array $fieldQueries;
+
+    /** @param list<FieldQueryInterface|Serializable|array|stdClass> $fieldQueries */
+    public function __construct(array $fieldQueries)
+    {
+        if (! array_is_list($fieldQueries)) {
+            throw new InvalidArgumentException('Expected filters to be a list, invalid array given.');
+        }
+
+        // Flatten nested CombinedFieldQuery
+        $this->fieldQueries = array_reduce($fieldQueries, static function (array $fieldQueries, mixed $fieldQuery): array {
+            if ($fieldQuery instanceof CombinedFieldQuery) {
+                return array_merge($fieldQueries, $fieldQuery->fieldQueries);
             }
+
+            $fieldQueries[] = $fieldQuery;
+
+            return $fieldQueries;
+        }, []);
+
+        // Validate FieldQuery types and non-duplicate operators
+        $seenOperators = [];
+        foreach ($this->fieldQueries as $fieldQuery) {
+            if ($fieldQuery instanceof FieldQueryInterface && $fieldQuery instanceof OperatorInterface) {
+                $operator = $fieldQuery->getOperator();
+            } elseif (is_array($fieldQuery)) {
+                if (count($fieldQuery) !== 1) {
+                    throw new InvalidArgumentException(sprintf('Operator array must contain exactly one key starting with $. %d given.', count($fieldQuery)));
+                }
+
+                $operator = array_key_first($fieldQuery);
+                if (! str_starts_with($operator, '$')) {
+                    throw new InvalidArgumentException(sprintf('Operator array must contain exactly one key starting with $. "%s" given.', $operator));
+                }
+            } elseif ($fieldQuery instanceof stdClass) {
+                $fieldQuery = get_object_vars($fieldQuery);
+                if (count($fieldQuery) !== 1) {
+                    throw new InvalidArgumentException(sprintf('Operator object must contain exactly one key starting with $. %d given.', count($fieldQuery)));
+                }
+
+                $operator = array_key_first($fieldQuery);
+                if (! str_starts_with($operator, '$')) {
+                    throw new InvalidArgumentException(sprintf('Operator object must contain exactly one key starting with $. "%s" given.', $operator));
+                }
+            } elseif ($fieldQuery instanceof Serializable) {
+                // Unknown operator, let the server handle it
+                continue;
+            } else {
+                throw new InvalidArgumentException(sprintf('Expected filters to be a list of field query operators, BSON documents, array or stdClass, %s given.', get_debug_type($fieldQuery)));
+            }
+
+            if (array_key_exists($operator, $seenOperators)) {
+                throw new InvalidArgumentException(sprintf('Duplicate operator "%s" detected.', $operator));
+            }
+
+            $seenOperators[$operator] = true;
         }
     }
 }
--- a/src/Builder/Type/CombinedFieldQuery.php
+++ b/src/Builder/Type/CombinedFieldQuery.php
@@ -16,11 +16,12 @@ use function count;
 use function get_debug_type;
 use function get_object_vars;
 use function is_array;
+use function is_string;
 use function sprintf;
 use function str_starts_with;
 
 /**
- * List of field query that apply to the same field path.
+ * List of field queries that apply to the same field path.
  */
 class CombinedFieldQuery implements FieldQueryInterface
 {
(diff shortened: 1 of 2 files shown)


src/Builder/Type/CombinedFieldQuery.php
```diff
@@ -4,28 +4,84 @@
...
+    public readonly array $fieldQueries;
+
+    /** @param list<FieldQueryInterface|Serializable|array|stdClass> $fieldQueries */
+    public function __construct(array $fieldQueries)
+    {
+        if (! array_is_list($fieldQueries)) {
+            throw new InvalidArgumentException('Expected filters to be a list, invalid array given.');
+        }
+
+        // Flatten nested CombinedFieldQuery
```
jmikola, line 41: This intentionally handles only one level of nesting, correct? Is there any use case where CombinedFieldQuery might be nested further, or if so perhaps it's not relevant to flatten in this constructor?

src/Builder/Type/CombinedFieldQuery.php
```diff
@@ -4,28 +4,84 @@
...
+        }, []);
+
+        // Validate FieldQuery types and non-duplicate operators
+        $seenOperators = [];
+        foreach ($this->fieldQueries as $fieldQuery) {
+            if ($fieldQuery instanceof FieldQueryInterface && $fieldQuery instanceof OperatorInterface) {
+                $operator = $fieldQuery->getOperator();
+            } elseif (is_array($fieldQuery)) {
+                if (count($fieldQuery) !== 1) {
+                    throw new InvalidArgumentException(sprintf('Operator array must contain exactly one key starting with $. %d given.', count($fieldQuery)));
```
jmikola: This shares the same exception message as the `str_starts_with` validation below, but is reporting the size of the array instead of an invalid string. I think it'd be clearer to just say "must contain exactly one key, %d given". Alternatively, say: "%d keys given" to make this clear.

I might also suggest quoting the "$" character in these messages for readability.

src/Builder/Type/CombinedFieldQuery.php
```diff
@@ -4,28 +4,84 @@
...
+                $fieldQuery = get_object_vars($fieldQuery);
+                if (count($fieldQuery) !== 1) {
+                    throw new InvalidArgumentException(sprintf('Operator object must contain exactly one key starting with $. %d given.', count($fieldQuery)));
+                }
+
+                $operator = array_key_first($fieldQuery);
+                if (! str_starts_with($operator, '$')) {
+                    throw new InvalidArgumentException(sprintf('Operator object must contain exactly one key starting with $. "%s" given.', $operator));
+                }
+            } elseif ($fieldQuery instanceof Serializable) {
```
jmikola: This is just checking for the `MongoDB\BSON\Serializable` interface. Did you mean to allow `MongoDB\BSON\Document` as well?

The exception below says the "BSON documents" are permitted.

src/Builder/Type/CombinedFieldQuery.php
```diff
@@ -4,28 +4,84 @@
...
+                // Unknown operator, let the server handle it
```
jmikola: Just to confirm: this isn't suggesting that the operator is invalid, just that you're choosing not to inspect it further, correct?

In PHPLIB, there are cases where I invoke `bsonSerialize()` to examine the Serializable object. I suppose that could be done here, but I understand if that'd be a waste of effort for such an edge case.

Do we actually expect users to pass operators as BSON Documents and Serializable instances to express _operators_? I could see that used for document values (e.g. ResolvesToDocument) but I'd be fine not supporting those edge cases in the builder for contexts such as this.

tests/Builder/Type/CombinedFieldQueryTest.php
```diff
@@ -21,9 +24,31 @@ public function testEmptyFieldQueries(): void
...
+            (object) ['$lt' => 1],
+            new class implements Serializable {
+                public function bsonSerialize(): array
+                {
+                    return ['$gte' => 1];
+                }
+            },
+        ]);
+
+        $this->assertCount(4, $fieldQueries->fieldQueries);
```
jmikola: I suppose asserting the output would have no effect as the constructor doesn't change the values in any way -- it just validates. Correct?

tests/Builder/Type/CombinedFieldQueryTest.php
```diff
@@ -44,5 +69,49 @@ public static function provideInvalidFieldQuery(): Generator
         yield 'string' => ['foo'];
         yield 'bool' => [true];
         yield 'null' => [null];
+        yield 'empy array' => [[]];
+        yield 'empy object' => [(object) []];
```
jmikola:
```suggestion
        yield 'empty array' => [[]];
        yield 'empty object' => [(object) []];
```

tests/Builder/Type/CombinedFieldQueryTest.php
```diff
@@ -44,5 +69,49 @@ public static function provideInvalidFieldQuery(): Generator
...
+
+    /**
+     * @param array<mixed> $fieldQueries
+     *
+     * @dataProvider provideDuplicateOperator
+     */
+    public function testRejectDuplicateOperator(array $fieldQueries): void
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('Duplicate operator "$eq" detected.');
```
jmikola: Suggestion to remove periods from single-sentence messages.

tests/Builder/Type/CombinedFieldQueryTest.php
```diff
@@ -44,5 +69,49 @@ public static function provideInvalidFieldQuery(): Generator
...
+                new GtOperator(2),
+            ],
+        ];
+
+        yield 'object and array' => [
+            [
+                (object) ['$ne' => 1],
+                ['$ne' => 2],
+            ],
+        ];
```
jmikola, line 111: I assume the case of a Serializable returning a duplicate operator is intentionally untested, since the current logic would consider it valid. Do you want to record that in a separate, passing test to demonstrate that Serializable objects are _not_ examined?

jmikola (commented):
Do you want to add a phpcs rule to suppress the `Superfluous suffix "Interface"` messages?

GromNaN:
> Do you want to add a phpcs rule to suppress the `Superfluous suffix "Interface"` messages?

Done in https://github.com/mongodb/mongo-php-builder/pull/4/files#diff-05ae9cddcaec1e845771a7db224961439f83ef5939ec67d3a48744cb34d7e58bR28

src/Builder/Type/CombinedFieldQuery.php
GromNaN: Yes. There can't be multiple levels, since this is automatically flattened when a CombinedFieldQuery is created.

src/Builder/Type/CombinedFieldQuery.php
GromNaN: Yes, I agree the type system should not accept BSON types for operators. I can fine-tune the accepted types for `query` and `fieldQuery`.

tests/Builder/Type/CombinedFieldQueryTest.php
GromNaN: Exact.

tests/Builder/Type/CombinedFieldQueryTest.php
GromNaN: Fixed.

tests/Builder/Type/CombinedFieldQueryTest.php
GromNaN: Done.

tests/Builder/Type/CombinedFieldQueryTest.php
GromNaN: I removed support for Serializable for now.

src/Builder/Type/CombinedFieldQuery.php
```diff
@@ -4,28 +4,80 @@
...
+use function count;
 use function get_debug_type;
+use function get_object_vars;
 use function is_array;
 use function sprintf;
+use function str_starts_with;
 
 /**
- * List of filters that apply to the same field path.
+ * List of field query that apply to the same field path.
```
alcaeus:
```suggestion
 * List of field queries that apply to the same field path.
```

src/Builder/Type/CombinedFieldQuery.php
```diff
@@ -4,28 +4,80 @@
...
+            } elseif ($fieldQuery instanceof stdClass) {
+                $fieldQuery = get_object_vars($fieldQuery);
+                if (count($fieldQuery) !== 1) {
+                    throw new InvalidArgumentException(sprintf('Operator object must contain exactly one key. %d given', count($fieldQuery)));
+                }
+
+                $operator = array_key_first($fieldQuery);
+                if (! str_starts_with($operator, '$')) {
+                    throw new InvalidArgumentException(sprintf('Operator object must contain exactly one key starting with $. "%s" given', $operator));
+                }
```
alcaeus: Suggestion: check for `stdClass` first and convert it to an array, which would allow you to use the same logic for both types.

src/Builder/Type/CombinedFieldQuery.php
```diff
@@ -4,28 +4,80 @@
...
+
+    /** @param list<FieldQueryInterface|array|stdClass> $fieldQueries */
+    public function __construct(array $fieldQueries)
+    {
+        if (! array_is_list($fieldQueries)) {
+            throw new InvalidArgumentException('Expected filters to be a list, invalid array given.');
+        }
+
+        // Flatten nested CombinedFieldQuery
+        $this->fieldQueries = array_reduce($fieldQueries, static function (array $fieldQueries, mixed $fieldQuery): array {
```
alcaeus: Using a `FieldQueryInterface|array|stdClass` type on the `$fieldQuery` parameter would change some `InvalidArgumentException` to type errors, which is appropriate in this case:
```suggestion
        $this->fieldQueries = array_reduce($fieldQueries, static function (array $fieldQueries, FieldQueryInterface|array|stdClass $fieldQuery): array {
```

alcaeus (commented):
Some minor feedback for optimisations, but the logic itself seems sound.

src/Builder/Type/CombinedFieldQuery.php
GromNaN: Done.

src/Builder/Type/CombinedFieldQuery.php
GromNaN: The type was actually a lot longer list.

PR #4, opened 2023-10-17
Fix Javascript type & other PHP types

GromNaN:


generator/config/expression/function.yaml
```diff
@@ -11,9 +11,10 @@ arguments:
     -
         name: body
         type:
-            - string
+            - javascript
```
jmikola, line 14: Does this type also allow `string`? If so, does that mean we accept strings everywhere a Javascript type is expected in the builder? I haven't reviewed all of the operators but I don't know off hand if a string is accepted everywhere that a code type is.

GromNaN: Yes, string is accepted. I could convert `string` to `MongoDB\BSON\Javascript` object everywhere they are expected.

PR #6, opened 2023-10-18
PHPLIB-1271 Add tests from the documentation

GromNaN:
- ~In generator config, create files `<operator>.tests.js` to store a list of expected pipelines.~
- ~Use a Node script to convert the JS to PHP array~

- Add `tests:` item operators config Yaml files.
- Generate a PHPUnit test class for each operator.


generator/composer.json
```diff
@@ -15,9 +15,10 @@
     "require": {
         "php": ">=8.1",
         "ext-mongodb": "^1.16.0",
-        "mongodb/mongodb": "^1.17.0@dev",
         "mongodb/builder": "@dev",
+        "mongodb/mongodb": "^1.17.0@dev",
         "nette/php-generator": "^4",
+        "nikic/php-parser": "^4.17",
```
GromNaN, line 21: Required to modify existing files with the generator.

tests/Builder/Stage/AddFieldsStageTest.php
```diff
@@ -0,0 +1,96 @@
...
+
+        $expected = $this->getExpectedAddingFieldsToAnEmbeddedDocument();
+
+        $this->assertSamePipeline($expected, $pipeline);
+    }
+
+    /** @see getExpectedUsingTwoAaddFieldsStages */
+    public function testUsingTwoAaddFieldsStages(): void
+    {
+        $this->markTestSkipped('$sum must accept arrayFieldPath and render it as a single value: https://jira.mongodb.org/browse/PHPLIB-1287');
```
GromNaN, line 30: This pipeline can't be implemented. Work on `$sum` operator is required. [PHPLIB-1287](https://jira.mongodb.org/browse/PHPLIB-1287)

generator/config/schema.json
```diff
@@ -167,6 +174,28 @@
...
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "name": {
+                    "type": "string"
+                },
+                "link": {
+                    "type": "string",
+                    "format": "uri",
+                    "qt-uri-protocols": [
```
jmikola: I can assume what this does, but where is `qt-uri-protocols` documented? It doesn't appear to be generic JSON schema syntax.

phpcs.xml.dist
```diff
@@ -16,6 +16,7 @@
 
     <!-- Ignore auto-generated code -->
     <exclude-pattern>src/Builder/(Accumulator|Expression|Query|Projection|Stage)/*\.php</exclude-pattern>
+    <exclude-pattern>*\.js</exclude-pattern>
```
jmikola: Why would JS files need to be excluded for `phpcs`?

tests/Builder/Accumulator/AccumulatorAccumulatorTest.php
```diff
@@ -0,0 +1,147 @@
...
+                    accumulate: new Javascript('function (state, numCopies) { return { count: state.count + 1, sum: state.sum + numCopies } }'),
+                    accumulateArgs: [Expression::fieldPath('copies')],
+                    merge: new Javascript('function (state1, state2) { return { count: state1.count + state2.count, sum: state1.sum + state2.sum } }'),
+                    finalize: new Javascript('function (state) { return (state.sum / state.count) }'),
+                    lang: 'js',
+                ),
+            ),
+        );
+
+        $expected = $this->getExpectedUseAccumulatorToImplementTheAvgOperator();
```
jmikola: If I understand correctly, only the `get` methods are generated and you still need to write the test yourself. Rather than embed generated methods _within_ non-generated _files_, why not just generate JSON files which are then referenced from these tests?

Alternatively, you can generate actual PHP files that return the value to compare against. I suppose that could be either an array or `MongoDB\BSON\PackedArray` instance.

tests/Builder/PipelineTestCase.php
```diff
@@ -0,0 +1,27 @@
...
+use MongoDB\Builder\Pipeline;
+use PHPUnit\Framework\TestCase;
+
+use function MongoDB\BSON\fromJSON;
+use function MongoDB\BSON\toPHP;
+use function var_export;
+
+class PipelineTestCase extends TestCase
+{
+    final public static function assertSamePipeline(string $expected, Pipeline $pipeline): void
```
jmikola: I assume this is the first use of this function. If we keep it as-is, renaming `$expected` to `$expectedJson` may be clearer. I do think it may be preferable to avoid JSON entirely and just generate PHP files that return a value that can be included here.

Those generated files could then have the appropriate "auto-generated" header and be ignored via `.gitattributes`.

phpcs.xml.dist
GromNaN: > By default, PHP_CodeSniffer will check any file it finds with a .inc, .php, .js or .css extension

https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#specifying-valid-file-extensions

GromNaN: I remove it, now that I removed the js script.

generator/config/schema.json
GromNaN: Thanks Copilot. It changed to use a pattern. That's [documented](https://json-schema.org/understanding-json-schema/reference/string#regexp).

generator/config/query/regex.yaml
```diff
@@ -11,3 +11,27 @@ arguments:
...
+        name: 'Perform a LIKE Match'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/regex/#perform-a-like-match'
+        pipeline:
+            -
+                $match:
+                    sku:
+                        # Should be nested in $regex
+                        $regularExpression:
+                            pattern: '789$'
+                            options: ''
```
GromNaN, line 26: `$regex` is a special key. I can't use it for a regular object.

```php
$json = <<<'JSON'
{
    "foo": {
        "$regex": {
            "$regularExpression": {
                "pattern": "789$",
                "options": "i"
            }
        }
    }
}
JSON;

$document = fromJSON($json);

// Fatal error: Uncaught MongoDB\Driver\Exception\UnexpectedValueException: Unexpected nested object value for "$regex" key 
```

jmikola: That exception originates from [`_bson_json_read_start_map()`](https://github.com/mongodb/mongo-c-driver/blob/1.24.4/src/libbson/src/bson/bson-json.c#L1215)

Also discussed in the Extended JSON spec:

 * [Parsers](https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#parsers)
 * [Of Canonical Type Wrapper Formats: RegularExpression](https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#regularexpression)
 * [Q&A](https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#q-a)

If you're using `$regex` as a query operator, the "Parsers" section above says:

> A parser that accepts Legacy Extended JSON MUST be configurable such that a JSON text of a MongoDB query filter containing the [regex](https://www.mongodb.com/docs/manual/reference/operator/query/regex/) query operator can be parsed.

And then it proceeds to give two examples. The first example resembles what you're trying to do here, where the `$regex` value is the extended JSON representation of a regex object (using the `$regularExpression` key). I'm not sure why that's prohibited in libbson, but it may be worth asking the C team about.

That aside, is anything preventing you from using the following?

```
{ $regex: 'pattern', $options: '<options>' }
```

GromNaN: > That aside, is anything preventing you from using the following?
> `{ $regex: 'pattern', $options: '<options>' }`

This will result in `new MongoDB\BSON\Regex('pattern', '<options'>)`.

The codec wraps the regex in the `$regex` operator for consistency with other operators.

The expected query that I need to encode in Yaml/Json is:
```
(object) ['sku' => (object) ['$regex' => new MongoDB\BSON\Regex('789$', '')]]
```

GromNaN:
```php
$query = (object) ['sku' => (object) ['$regex' => new MongoDB\BSON\Regex('789$', '')]];

$json = MongoDB\BSON\toJSON(MongoDB\BSON\fromPHP($x));

// { "sku" : { "$regex" : { "$regex" : "789$", "$options" : "" } } }

MongoDB\BSON\fromJSON($json);

// Warning: Uncaught MongoDB\Driver\Exception\UnexpectedValueException: Unexpected nested object value for "$regex" key in php shell code:1
```

alcaeus (commented):
Changes LGTM. No objection to using JSON for pipelines until we're able to read JSON pipelines into BSON without encountering parsing issues. 👍

generator/config/query/regex.yaml
GromNaN: Issue solved by comparing JSON: https://github.com/mongodb/mongo-php-builder/pull/13

PR #8, opened 2023-10-20
Add test for `$unwind` and a complex `$group` stage

OskarStark:
I am using it this way in my pipeline and when I use the `assertSamePipeline` method from this repository (copy/pasted), I receive the following error:
CleanShot 2023-10-20 at 21 56 07

With this test in your CI I want to achieve, that it is not a problem of the `assertSamePipeline` method.

cc @alcaeus @GromNaN

### DIFF
--- a/tests/Builder/BuilderEncoderTest.php
+++ b/tests/Builder/BuilderEncoderTest.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace MongoDB\Tests\Builder;
 
-use App\Import\Value\Field\Fieldtype;
 use Generator;
 use MongoDB\Builder\Accumulator;
 use MongoDB\Builder\BuilderEncoder;
--- a/tests/Builder/BuilderEncoderTest.php
+++ b/tests/Builder/BuilderEncoderTest.php
@@ -287,8 +287,45 @@ class BuilderEncoderTest extends TestCase
         );
 
         $expected = [
-            ['$unwind' => '$fielgroups'],
-            ['$unwind' => '$fielgroups.fields'],
+            ['$unwind' => '$fieldgroups'],
+            ['$unwind' => '$fieldgroups.fields'],
+        ];
+
+        $this->assertSamePipeline($expected, $pipeline);
+    }
+
+    public function testComplexGroupStage(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::group(
+                _id: null,
+                results: Accumulator::push(
+                    Expression::arrayToObject([[
+                       'k' => Expression::toString(Expression::fieldPath('key')),
+                       'v' => Expression::fieldPath('value'),
+                    ]]),
+                ),
+            ),
+        );
+
+        $expected = [
+            [
+                '$group' => [
+                    '_id' => null,
+                    'results' => [
+                        '$push' => [
+                            '$arrayToObject' => [
+                                [
+                                    [
+                                        'k' => ['$toString' => '$key'],
+                                        'v' => '$value',
+                                    ],
+                                ],
+                            ],
+                        ],
+                    ],
+                ],
+            ],
         ];
 
         $this->assertSamePipeline($expected, $pipeline);
--- a/tests/Builder/BuilderEncoderTest.php
+++ b/tests/Builder/BuilderEncoderTest.php
@@ -282,13 +282,15 @@ class BuilderEncoderTest extends TestCase
     public function testUnwind(): void
     {
         $pipeline = new Pipeline(
-            Stage::unwind('$fieldgroups'),
-            Stage::unwind('$fieldgroups.fields'),
+            Stage::unwind('$foo'),
+            Stage::unwind('$foo.bar'),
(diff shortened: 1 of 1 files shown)


tests/Builder/BuilderEncoderTest.php
```diff
@@ -4,6 +4,7 @@
 
 namespace MongoDB\Tests\Builder;
 
+use App\Import\Value\Field\Fieldtype;
```
OskarStark:
```suggestion
```

OskarStark:
I rewrote all my pipelines, and my integration tests are all passing, except the ones, where I compare the pipeline itself. That means, the pipeline is correct, the fetched data is correct, but the assertion method seems not to be able to handle `$unwind` and my complex `$group` stage

tests/Builder/BuilderEncoderTest.php
```diff
@@ -279,6 +279,60 @@ public function testRedactStage(): void
...
+    public function testUnwind(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::unwind('$foo'),
+            Stage::unwind('$foo.bar'),
+            Stage::unwind(Expression::arrayFieldPath('foo.bar.baz')),
+        );
+
+        $expected = [
+            ['$unwind' => '$foo'],
```
GromNaN, line 291: This is the short syntax of [`$unwind`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/unwind/). In order to cover all use cases while keeping the code as simple as possible, I've chosen not to detect when only the field path is provided. So it's always an object with at least one property "path" that's returned.

```suggestion
            ['$unwind' => ['path' => '$foo']],
```

tests/Builder/BuilderEncoderTest.php
```diff
@@ -279,6 +279,60 @@ public function testRedactStage(): void
...
+    public function testComplexGroupStage(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::group(
+                _id: null,
+                results: Accumulator::push(
+                    Expression::arrayToObject([[
+                       'k' => Expression::toString(Expression::fieldPath('key')),
+                       'v' => Expression::fieldPath('value'),
+                    ]]),
```
GromNaN, line 308: It's strictly equivalent. But in order to match the expected pipeline, it have to be an object. Maybe `assertSamePipeline` need to be improved to accept array or object equally.
```suggestion
                    Expression::arrayToObject([
                        object(
                            k: Expression::toString(Expression::fieldPath('key')),
                            v: Expression::fieldPath('value'),
                        ),
                    ]),
```

GromNaN (commented):
Thanks for your feedbacks. 
MongoDB and the PHP driver are fairly permissive about the types they accept. We need to choose the most reliable for the encoder to cover all use cases.

tests/Builder/BuilderEncoderTest.php
OskarStark: Ok it would be helpful if the diff would show this when failing

tests/Builder/BuilderEncoderTest.php
OskarStark: Yes it should not be necessary to change the pipeline itself to make the assert method happy 👌🏻

OskarStark:
I changed my pipelines accordingly and my tests pass 👍 Thanks for the help

GromNaN:
I'm closing your PR for now. The requested changes are tracked in [PHPLIB-1291](https://jira.mongodb.org/browse/PHPLIB-1291).

tests/Builder/BuilderEncoderTest.php
GromNaN: Using JSON comparison, the error of looks like this:

```diff
1) MongoDB\Tests\Builder\BuilderEncoderTest::testUnwind
'{ "root" : [ { "$unwind" : { "path" : "$foo" } }, { "$unwind" : { "path" : "$foo.bar" } }, { "$unwind" : { "path" : "$foo.bar.baz" } } ] }'
Failed asserting that '{ "root" : [ { "$unwind" : { "path" : "$foo" } }, { "$unwind" : { "path" : "$foo.bar" } }, { "$unwind" : { "path" : "$foo.bar.baz" } } ] }' matches JSON string "{ "root" : [ { "$unwind" : "$foo" }, { "$unwind" : "$foo.bar" }, { "$unwind" : "$foo.bar.baz" } ] }".
--- Expected
+++ Actual
@@ @@
 {
     "root": [
         {
-            "$unwind": "$foo"
+            "$unwind": {
+                "path": "$foo"
+            }
         },
         {
-            "$unwind": "$foo.bar"
+            "$unwind": {
+                "path": "$foo.bar"
+            }
         },
         {
-            "$unwind": "$foo.bar.baz"
+            "$unwind": {
+                "path": "$foo.bar.baz"
+            }
         }
     ]
 }

```

tests/Builder/BuilderEncoderTest.php
GromNaN: This is solved by using a BSON round-trip. See https://github.com/mongodb/mongo-php-builder/pull/13

PR #9, opened 2023-10-30
PHPLIB-1292 Accept a single array as query to support filter on numerical field names

GromNaN:
Fix [PHPLIB-1292](https://jira.mongodb.org/browse/PHPLIB-1292)

Numerical field names can't be used with named argument, because even if they comes from an array unpacking, the numerical keys are recalculated as a sequence: https://3v4l.org/Gj7jC

Everywhere a query is expected, an single array can be passed instead of the variadic arguments.

```php
Stage::match(['1' => Query::eq(2)]);
Stage::match(...['1' => Query::eq(2)]); // Incorrect
```

Identicals:
```php
Stage::match(['foo' => Query::eq(2)]);
Stage::match(foo: Query::eq(2));
```

### DIFF
--- a/generator/config/expressions.php
+++ b/generator/config/expressions.php
@@ -83,11 +83,11 @@ return $expressions + [
     ],
     'fieldQuery' => [
         'returnType' => Type\FieldQueryInterface::class,
-        'acceptedTypes' => [Type\FieldQueryInterface::class, ...$bsonTypes['any']],
+        'acceptedTypes' => [Type\FieldQueryInterface::class, BSON\Regex::class, 'bool', 'int', 'float', 'string', 'array', 'null', stdClass::class],
     ],
     'query' => [
         'returnType' => Type\QueryInterface::class,
-        'acceptedTypes' => [Type\QueryInterface::class, ...$bsonTypes['object']],
+        'acceptedTypes' => [Type\QueryInterface::class, 'array'],
     ],
     'projection' => [
         'returnType' => Type\ProjectionInterface::class,
--- a/generator/config/expressions.php
+++ b/generator/config/expressions.php
@@ -83,7 +83,7 @@ return $expressions + [
     ],
     'fieldQuery' => [
         'returnType' => Type\FieldQueryInterface::class,
-        'acceptedTypes' => [Type\FieldQueryInterface::class, BSON\Regex::class, 'bool', 'int', 'float', 'string', 'array', 'null', stdClass::class],
+        'acceptedTypes' => [Type\FieldQueryInterface::class, BSON\Regex::class, 'bool', 'string', 'array', 'null', stdClass::class, ...$bsonTypes['number']],
     ],
     'query' => [
         'returnType' => Type\QueryInterface::class,
--- a/generator/config/query/and.yaml
+++ b/generator/config/query/and.yaml
@@ -12,4 +12,4 @@ arguments:
         type:
             - query
         variadic: array
-        variadicMin: 2
+        variadicMin: 1
--- a/generator/config/query/and.yaml
+++ b/generator/config/query/and.yaml
@@ -8,7 +8,8 @@ description: |
     Joins query clauses with a logical AND returns all documents that match the conditions of both clauses.
 arguments:
     -
-        name: expression
+        name: queries
         type:
             - query
         variadic: array
+        variadicMin: 2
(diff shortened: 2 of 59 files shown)


generator/config/query/and.yaml
```diff
@@ -8,7 +8,8 @@ description: |
     Joins query clauses with a logical AND returns all documents that match the conditions of both clauses.
 arguments:
     -
-        name: expression
+        name: queries
         type:
             - query
         variadic: array
+        variadicMin: 2
```
alcaeus: Why `variadicMin: 2`? While `$and` rejects an empty list on the server side, it works just fine with a single query. Either way, I'm not sure we need to check for this at all. Note this applies to `$or` and `$nor` as well:
```
Enterprise test> db.test.find({$and: []})
MongoServerError: $and/$or/$nor must be a nonempty array
Enterprise test> db.test.find({$and: [{foo: { $exists: false }}]})
// No error
```

alcaeus (commented):
Minor issue with `variadicMin`, but LGTM provided that the named argument syntax continues to work.

generator/config/query/and.yaml
GromNaN: I have this feature for checking that is used in other places. I'll reduce to 1.

GromNaN:
This PR has been subject to inflation.
- Tuned the accepted types for field query
- Fixed CS (newline after if) in generated code

generator/config/expressions.php
```diff
@@ -83,11 +83,11 @@
     ],
     'fieldQuery' => [
         'returnType' => Type\FieldQueryInterface::class,
-        'acceptedTypes' => [Type\FieldQueryInterface::class, ...$bsonTypes['any']],
+        'acceptedTypes' => [Type\FieldQueryInterface::class, BSON\Regex::class, 'bool', 'string', 'array', 'null', stdClass::class, ...$bsonTypes['number']],
```
jmikola, line 86: Why is Regex specifically allowed but not other special BSON types? Does everything else require use of `eq()` or some other comparison operator?

src/Builder/Type/OutputWindow.php
```diff
@@ -62,7 +62,7 @@ public function __construct(
                 throw new InvalidArgumentException(sprintf('Expected $documents argument to be a list of 2 string or int. Got [%s, %s]', get_debug_type($documents[0]), get_debug_type($documents[1])));
             }
 
-            $window ??= new stdClass();
+            $window = new stdClass();
```
jmikola, line 65: Did static analysis complain about this? I assume `??=` was there because it was more flexible if a subsequent change resulted in this conditional no longer appearing first.

tests/Builder/Type/QueryObjectTest.php
```diff
@@ -51,19 +66,23 @@ public function provideQueryObjectValue(): Generator
...
         yield 'query operator' => [[new CommentOperator('foo'), 'foo' => 1], 2];
+        yield 'numeric field with array' => [[1 => [2, 3]]];
+        yield 'numeric field with operator' => [[1 => ['$eq' => 2]]];
     }
 
     public function testFieldQueryList(): void
     {
         $queryObject = QueryObject::create(
-            foo: [new GtOperator(1), new LtOperator(5)],
+            ['foo' => [new GtOperator(1), new LtOperator(5)]],
```
jmikola, line 85: In an earlier comment, @alcaeus wrote:

> LGTM provided that the named argument syntax continues to work

Are you still testing with named arg syntax elsewhere?

generator/src/OperatorClassGenerator.php
```diff
@@ -76,11 +76,12 @@ public function createClass(GeneratorDefinition $definition, OperatorDefinition
                 $constuctor->setVariadic();
                 $constuctor->addComment('@param ' . $type->doc . ' ...$' . $argument->name . rtrim(' ' . $argument->description));
 
-                if ($argument->variadicMin !== null) {
+                if ($argument->variadicMin > 0) {
```
jmikola, line 79: Noted that ArgumentDefinition always sets this as an integer value, and `schema.json` likewise enforces only integer values if the config option is defined at all.

src/Builder/Type/OutputWindow.php
GromNaN: Yes, static analysis.

tests/Builder/Type/QueryObjectTest.php
GromNaN: Yes, but not for this internal constructor. The variadic named arguments is supported by `MongoDB\Builder\Query::query(...$queries)`

GromNaN: Tests on `MongoDB\Builder\Query::query(...$queries)` are only in the integration tests (BuilderEncoderTests and various operator tests).

generator/config/expressions.php
GromNaN: Reverted in https://github.com/mongodb/mongo-php-builder/pull/11

PR #12, opened 2023-11-03
Support using first, firstN, last, lastN accumulators as expressions

alcaeus:
As per the documentation, the `$first`, `$firstN`, `$last`, and `$lastN` accumulators can also be used as aggregation expressions within non-grouping stages, e.g. `$addFields`. This PR makes sure these can be used accordingly.

I noted that other operators, such as `$max` and `$sum` have been duplicated as expressions, but those operators have a different syntax depending on whether they are used as expressions or accumulators. Since these array accumulators always have the same syntax, I decided to only leave them in as accumulators and avoid duplicating the config file in the generator. @GromNaN let me know if you'd rather duplicate the config to also add factories for these operators in the `Expression` builder class.

### DIFF
--- a/generator/config/accumulator/first.yaml
+++ b/generator/config/accumulator/first.yaml
@@ -4,6 +4,7 @@ link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/first/
 type:
     - accumulator
     - window
+    - resolvesToAny
 encode: single
 description: |
     Returns the result of an expression for the first document in a group or window.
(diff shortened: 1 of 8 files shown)


GromNaN:
Indeed, I missed this point when I created the `Accumulator` namespace.
According to the current design, they should be duplicated, but from a user point of view it's not ideal. 
In fact, I think that grouping into Stage/Accumulator/Window/Expression should be done according to the return type, not a fixed classification. We could then repeat the same factories method in different factory classes.

GromNaN:
Replaced by https://github.com/mongodb/mongo-php-builder/pull/14

PR #14, opened 2023-11-03
Add `first`, `firstN`, `last`, `lastN` operators and tests

GromNaN:
Replace https://github.com/mongodb/mongo-php-builder/pull/12

This operators are accumulators and expression operators.

### DIFF
--- a/generator/config/accumulator/first.yaml
+++ b/generator/config/accumulator/first.yaml
@@ -13,3 +13,46 @@ arguments:
         name: expression
         type:
             - expression
+tests:
+    -
+        name: 'Use in $group Stage'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/first/#use-in--group-stage'
+        pipeline:
+            -
+                $sort:
+                    item: 1
+                    date: 1
+            -
+                $group:
+                    _id: '$item'
+                    firstSale:
+                        $first: '$date'
+    -
+        name: 'Missing Data'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/first/#missing-data'
+        pipeline:
+            -
+                $sort:
+                    item: 1
+                    price: 1
+            -
+                $group:
+                    _id: '$item'
+                    inStock:
+                        $first: '$quantity'
+    -
+        name: 'Use in $setWindowFields Stage'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/first/#use-in--setwindowfields-stage'
+        pipeline:
+            -
+                $setWindowFields:
+                    partitionBy: '$state'
+                    sortBy:
+                        orderDate: 1
+                    output:
+                        firstOrderTypeForState:
+                            $first: '$type'
+                            window:
+                                documents:
+                                    - 'unbounded'
+                                    - 'current'
--- a/generator/config/accumulator/first.yaml
+++ b/generator/config/accumulator/first.yaml
@@ -28,19 +28,6 @@ tests:
                     firstSale:
                         $first: '$date'
     -
-        name: 'Missing Data'
-        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/first/#missing-data'
-        pipeline:
-            -
-                $sort:
-                    item: 1
-                    price: 1
-            -
-                $group:
-                    _id: '$item'
-                    inStock:
-                        $first: '$quantity'
-    -
         name: 'Use in $setWindowFields Stage'
         link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/first/#use-in--setwindowfields-stage'
         pipeline:
--- a/generator/config/accumulator/lastN.yaml
+++ b/generator/config/accumulator/lastN.yaml
@@ -2,10 +2,13 @@
 name: $lastN
 link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/lastN/'
 type:
+    - accumulator
     - window
 encode: object
 description: |
-    Returns a specified number of elements from the end of an array. Distinct from the $lastN accumulator.
+    Returns an aggregation of the last n elements within a group.
+    The elements returned are meaningful only if in a specified sort order.
+    If the group contains fewer than n elements, $lastN returns all elements in the group.
 arguments:
     -
         name: input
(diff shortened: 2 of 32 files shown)


generator/config/expression/first.yaml
```diff
@@ -0,0 +1,13 @@
...
+type:
+    - resolvesToAny
+encode: single
+description: |
+    Returns the result of an expression for the first document in an array.
+arguments:
+    -
+        name: expression
+        type:
+            - resolvesToArray
```
alcaeus, line 13:
```suggestion
            - resolvesToArray
tests:
    -
        name: 'Use in $addFields Stage'
        # Note: no link as this is not an official example
        pipeline:
            -
                $addFields:
                    firstItem:
                        $first: '$items'
```

generator/config/expression/last.yaml
```diff
@@ -0,0 +1,13 @@
...
+type:
+    - resolvesToAny
+encode: single
+description: |
+    Returns the result of an expression for the last document in an array.
+arguments:
+    -
+        name: expression
+        type:
+            - resolvesToArray
```
alcaeus, line 13:
```suggestion
            - resolvesToArray
tests:
    -
        name: 'Use in $addFields Stage'
        # Note: no link as this is not an official example
        pipeline:
            -
                $addFields:
                    lastItem:
                        $last: '$items'
```

generator/config/stage/setWindowFields.yaml
```diff
@@ -27,10 +27,3 @@ arguments:
         description: |
             Specifies the field(s) to append to the documents in the output returned by the $setWindowFields stage. Each field is set to the result returned by the window operator.
             A field can contain dots to specify embedded document fields and array fields. The semantics for the embedded document dotted notation in the $setWindowFields stage are the same as the $addFields and $set stages.
-    -
-        name: window
```
alcaeus, line 31: I double-checked the docs after this was removed. It's a little confusing, as the `window` option (and its nested `document` and `range` fields) are documented in the window stage docs, but seem to be part of an output field.

alcaeus (commented):
Suggested a couple of tests for `$first` and `$last` in `$addFields`, but the changes LGTM.

generator/config/stage/setWindowFields.yaml
GromNaN: It should be documented as `output.$.window`, `output.$.window.documents`...

generator/config/expression/first.yaml
GromNaN: Thanks for the test. I removed the note as the lack of `link` will the marked that this is not an example from the documentation.

generator/config/accumulator/first.yaml
```diff
@@ -13,3 +13,46 @@ arguments:
...
+        pipeline:
+            -
+                $sort:
+                    item: 1
+                    price: 1
+            -
+                $group:
+                    _id: '$item'
+                    inStock:
+                        $first: '$quantity'
```
GromNaN: This example is almost the same as the previous one. It makes sense in the documentation, but should we port all the examples into tests?

generator/config/expression/first.yaml
alcaeus: I wasn't sure whether the generator requires a link, so I wanted to pointed out in the suggestion. It's not required in the final product at all.

generator/config/accumulator/first.yaml
alcaeus: I think the example makes sense in the docs to show behaviour of a missing field, but it's probably not relevant here. I'm fine removing it from the config, unless we want to make a point of having a builder version of each example in the documentation.

tests/Builder/Expression/Pipelines.php
```diff
@@ -10,4 +10,115 @@
...
+                        },
+                        "input": "$score"
+                    }
+                }
+            }
+        }
+    ]
+    JSON;
+
+    /** Use in $addFields Stage */
```
jmikola, line 48: Where are these comments coming from, and why do some of the cases in this enum have entirely different doc block comments? For instance, the singular "Example" descriptions look redundant.

generator/config/accumulator/first.yaml
jmikola: OK to remove this. I don't see the point in including examples with no benefit to testing.

generator/config/accumulator/lastN.yaml
```diff
@@ -2,10 +2,13 @@
 name: $lastN
 link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/lastN/'
 type:
+    - accumulator
     - window
```
jmikola, line 6: For [`$addToSet`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/addToSet/#mongodb-group-grp.-addToSet), I see the following clearly mentioned at the top of the page in the list of places where the operator is supported:

> * [$setWindowFields](https://www.mongodb.com/docs/manual/reference/operator/aggregation/setWindowFields/#mongodb-pipeline-pipe.-setWindowFields) (Available starting in MongoDB 5.0)

For [`$lastN`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/lastN/#window-function-and-aggregation-expression-support), the same information is buried much further down the page.

This seems very inconvenient.

----

Do all of the window operators have identical syntax to the non-window accumulator? I assume that's also why these configs appear under an `accumulator/` directory instead of being duplicated into a `window/` directory (as you had to do with `expression/`).

generator/config/stage/project.yaml
```diff
@@ -10,9 +10,6 @@ arguments:
     -
         name: specification
         type:
-            - bool
-            - int
+            - expression
```
jmikola, line 13: How is this change related to the operators in the PR title?

Is it correct to include `expression` here? Looking at the [`$project` syntax](https://www.mongodb.com/docs/manual/reference/operator/aggregation/project/#syntax), an expression is only supported as a _value_ for a field. Or am I misunderstanding and `type` here is only applying to the values for the keys/properties within the specification document?

generator/config/stage/setWindowFields.yaml
jmikola: `output` is not modeled at all, since it's an `object` type above. Does this leave us no way to utilize a builder for `window` and its sub-fields?

tests/Builder/Accumulator/FirstAccumulatorTest.php
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Test $first accumulator
+ */
+class FirstAccumulatorTest extends PipelineTestCase
+{
+    public function testMissingData(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::sort(object(
+                item: 1, // @todo accept true
```
jmikola: Is this TODO tracked anywhere? [PHPORM-105](https://jira.mongodb.org/browse/PHPORM-105) is the only sort-related ticket I found.

generator/config/stage/project.yaml
GromNaN: In the [`$lastN` example](https://www.mongodb.com/docs/manual/reference/operator/aggregation/lastN/#using--lastn-as-an-aggregation-expression), the operator is used in the `$project` stage. That's why I had to fix it.
You can see how it's used in [`Expression\LastNOperatorTest::testUsingLastNAsAnAggregationExpression`](https://github.com/mongodb/mongo-php-builder/pull/14/files#diff-d5bb1a8a1c567e077656e5f8948ef77bee8f0c04ca65775bf9586a913f45ed05R37)

generator/config/stage/setWindowFields.yaml
GromNaN: It's modeled using `MongoDB\Builder\Accumulator::window()` function.

tests/Builder/Accumulator/FirstAccumulatorTest.php
GromNaN: I don't have created an issue for all operators, we will have to check the behavior of all operators when implementing tests.

tests/Builder/Expression/Pipelines.php
GromNaN: I just stupidly copy-pasted the examples from the docs. We can improve each 
description as needed.

The one you are pointing isn't from the doc, that's why there is not link.

generator/config/accumulator/lastN.yaml
GromNaN: Window Operators are a subset of Group Accumulators. That's why I didn't create a separate category.

generator/config/stage/project.yaml
jmikola: Given that, `type` is referring to the _values_ of the `specification` object. I originally misunderstood and assumed this syntax would allow `specification` to be an `expression` directly.

jmikola (approved):
I think all of my questions are addressed.

generator/config/stage/project.yaml
GromNaN: Yes, that's the nuance with variadic arguments. The type applies to the values below the root.

PR #16, opened 2023-11-15
Reduce CI matrix as we don't use MongoDB server

GromNaN:
We don't have to test all the MongoDB versions and topologies for now, as we don't use the MongoDB client at all.


generator/composer.json
```diff
@@ -14,8 +14,7 @@
     },
     "require": {
         "php": ">=8.1",
-        "ext-mongodb": "^1.16.0",
-        "mongodb/builder": "@dev",
+        "ext-mongodb": "^1.17.0",
         "mongodb/mongodb": "^1.17.0@dev",
```
alcaeus: Might as well hold this pull request for a few more minutes until 1.17.0 is tagged ^^

alcaeus (approved):

generator/composer.json
GromNaN: Like this?
```suggestion
        "mongodb/mongodb": "^1.17.0",
```

PR #17, opened 2023-11-16
Add `$all` and `$elemMatch` field query operator tests

GromNaN:
## `$all`

It accepts values or field query operators.

## `$elemMatch`

It can be used as a query operator or as a field query operator.

### DIFF
--- a/generator/config/projection/elemMatch.yaml
+++ b/generator/config/projection/elemMatch.yaml
@@ -3,7 +3,7 @@ name: $elemMatch
 link: 'https://www.mongodb.com/docs/manual/reference/operator/projection/elemMatch/'
 type:
     - projection
-encode: object
+encode: single
 description: |
     Projects the first element in an array that matches the specified $elemMatch condition.
 arguments:
--- a/generator/config/query/all.yaml
+++ b/generator/config/query/all.yaml
@@ -10,5 +10,34 @@ arguments:
     -
         name: value
         type:
-            - any
+            - fieldQuery
         variadic: array
+tests:
+    -
+        name: 'Use $all to Match Values'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/all/#use--all-to-match-values'
+        pipeline:
+            -
+                $match:
+                    tags:
+                        $all:
+                            - 'appliance'
+                            - 'school'
+                            - 'book'
+    -
+        name: 'Use $all with $elemMatch'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/all/#use--all-with--elemmatch'
+        pipeline:
+            -
+                $match:
+                    qty:
+                        $all:
+                            -
+                                $elemMatch:
+                                    size: 'M'
+                                    num:
+                                        $gt: 50
+                            -
+                                $elemMatch:
+                                    num: 100
+                                    color: 'green'
--- a/src/Builder/Query.php
+++ b/src/Builder/Query.php
@@ -6,13 +6,16 @@ namespace MongoDB\Builder;
 
 use MongoDB\BSON\Regex;
 use MongoDB\BSON\Type;
+use MongoDB\Builder\Query\ElemMatchOperator;
 use MongoDB\Builder\Query\RegexOperator;
+use MongoDB\Builder\Type\CombinedFieldQuery;
 use MongoDB\Builder\Type\FieldQueryInterface;
 use MongoDB\Builder\Type\QueryInterface;
 use MongoDB\Builder\Type\QueryObject;
 use MongoDB\Exception\InvalidArgumentException;
 use stdClass;
 
+use function array_is_list;
 use function is_string;
 
 /**
@@ -24,6 +27,7 @@ final class Query
 {
     use Query\FactoryTrait {
         regex as private generatedRegex;
+        elemMatch as private generatedElemMatch;
     }
 
     /**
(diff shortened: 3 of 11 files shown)


generator/config/projection/elemMatch.yaml
```diff
@@ -3,7 +3,7 @@ name: $elemMatch
 link: 'https://www.mongodb.com/docs/manual/reference/operator/projection/elemMatch/'
 type:
     - projection
-encode: object
+encode: single
```
jmikola, line 6: Why was `object` used previously? I assume that would have resulted in incorrect nesting with an additional `query` property (based on the public property in the operator class) between `$elemMatch` and the actual value. Is that correct?

generator/config/query/all.yaml
```diff
@@ -11,4 +11,34 @@ arguments:
         name: value
         type:
             - any
+            - fieldQuery
```
jmikola, line 13: Is it redundant to specify `any` alongside `fieldQuery`? Looking at `generator/config/expressions.php`, I see the case for `fieldQuery` utilizes the following:

```
'acceptedTypes' => [Type\FieldQueryInterface::class, ...$bsonTypes['any']],
```

src/Builder/Query.php
```diff
@@ -40,6 +44,21 @@ public static function regex(Regex|string $regex, string|null $flags = null): Re
...
     }
 
+    /**
+     * The $elemMatch operator matches documents that contain an array field with at least one element that matches all the specified query criteria.
+     *
+     * @see https://www.mongodb.com/docs/manual/reference/operator/query/elemMatch/
+     */
+    public static function elemMatch(QueryInterface|FieldQueryInterface|Type|stdClass|array|bool|float|int|string|null ...$query): ElemMatchOperator
+    {
+        // $elemMatch accept a query or a field query
```
jmikola:
```suggestion
        // $elemMatch accepts a query or a field query
```

src/Builder/Query.php
```diff
@@ -40,6 +44,21 @@ public static function regex(Regex|string $regex, string|null $flags = null): Re
...
+        if (array_is_list($query)) {
+            return self::generatedElemMatch(new CombinedFieldQuery($query));
+        }
```
jmikola: What syntax is this referring to? It's not clear to me from looking at [$elemMatch: Syntax](https://www.mongodb.com/docs/manual/reference/operator/query/elemMatch/#syntax) and the [Single Query Condition](https://www.mongodb.com/docs/manual/reference/operator/query/elemMatch/#std-label-elemmatch-single-query-condition) section linked therein. IIRC, the value for an `$elemMatch` is always just a query object.

generator/config/projection/elemMatch.yaml
GromNaN: At first, it looked like an object with properties being each query operator. But since I introduced the Query object, this is the single value.
You guessed it, with encode=object, we had an exported property named "query".

generator/config/query/all.yaml
GromNaN: Indeed, reading the doc I hadn't identified that scalars were used as the implicit `$eq` here. Fixed.

src/Builder/Query.php
GromNaN: The `CombinedFieldQuery` is to support the [Element Match](https://www.mongodb.com/docs/manual/reference/operator/query/elemMatch/#element-match) example. There is no field name in the `$elemMatch` value :
```ts
{
  results: {
    $elemMatch: {
      $gte: 80,
      $lt: 85
  }
}
```

Without this code, the generated pipeline looks like this:
```ts
{
  results: {
    $elemMatch: {
      "0": { $gte: 80 },
      "1": { $lt: 85 }
    }
  }
}
```

I added a condition so that we accept a `QueryInterface` object if someone needs to use `0` as field name.

jmikola: Ah, for matching an element in an array of non-document values. That makes sense.

I think that'd be helpful to mention in the comment here (or maybe refer to "element match").

jmikola (approved):
LGTM but I think the comment for `$elemMatch` handling an "element match" would be improved with some more detail.

src/Builder/Query.php
```diff
@@ -40,6 +44,22 @@ public static function regex(Regex|string $regex, string|null $flags = null): Re
...
     }
 
+    /**
+     * The $elemMatch operator matches documents that contain an array field with at least one element that matches all the specified query criteria.
+     *
+     * @see https://www.mongodb.com/docs/manual/reference/operator/query/elemMatch/
+     */
+    public static function elemMatch(QueryInterface|FieldQueryInterface|Type|stdClass|array|bool|float|int|string|null ...$query): ElemMatchOperator
+    {
+        // $elemMatch accepts field query without field name ...
```
GromNaN: @jmikola what do you thing of this comment:
```suggestion
        // To match elements in non-document arrays, $elemMatch can be used with field query operators,
        // They must be passed as a list of operators without a field name as key.
```

GromNaN:
`$elemMatch` is both a query operator and a field query operator. I don't know if there is other operators like this. I need to handle that in `BuilderEncoder::encodeQueryObject()`.

src/Builder/Query.php
jmikola: Consider:

> $elemMatch can be used to match arrays of documents and non-document values alike. When matching non-document values, field query operators should be passed as a list of operators without the field name as a key.

Alternatively, you can replace `field query operators` with the class/interface name.

I assume this means users _cannot_ pass an associative array of the raw query operators, since that might be interpreted as the API for matching an array of documents?

```
[ '$gt' => 5, '$lt' => 10 ]
```

Might that be interpreted as a equality matches on "$gt" and "$lt" fields paths?

GromNaN: You're right.
I'm afraid that given the ambiguity, we have no choice but to remove the named arguments and make the use of a single `Query` or `FieldQuery` argument, which will be created with the `Query::query` or `Query::fieldQuery` function.

PR #18, opened 2023-11-16
PHPLIB-1273 Add `$switch` expression operator tests

GromNaN:
We need to add a model factory for each cases.

### DIFF
--- a/generator/config/expression/switch.yaml
+++ b/generator/config/expression/switch.yaml
@@ -24,3 +24,47 @@ arguments:
         description: |
             The path to take if no branch case expression evaluates to true.
             Although optional, if default is unspecified and no branch case evaluates to true, $switch returns an error.
+tests:
+    -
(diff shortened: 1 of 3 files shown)


generator/config/expression/switch.yaml
```diff
@@ -24,3 +24,47 @@ arguments:
...
+                                            - 90
+                                    then: 'Doing great!'
+                                -
+                                    case:
+                                        $and:
+                                            -
+                                                $gte:
+                                                    -
+                                                        $avg:
+                                                            - '$scores'
```
GromNaN, line 53: On the [example](https://www.mongodb.com/docs/manual/reference/operator/aggregation/switch/), this is not an array. But the `$avg` operator accepts both a single value or a list of values, and that behave the same way.

```suggestion
                                                        $avg: '$scores'
```

PR #19, opened 2023-12-21
Various improvements

alcaeus:
This PR adds a few improvements I've noticed while working on some pipelines:
* `$mergeObjects` is also available as an expression operator. I've followed the same strategy we previously used for `$first`
* The `defaults` argument for `$zip` should be optional
* `$add` and `$subtract` only implemented `ResolvesToNumber`. This made them unsuitable for use in places where an integer was required (in my case the `n` argument to `$slice`). I've changed the operators to resolve to all numeric types individually. Note that the pipeline may error on the server if `$add` evaluates to a value of the wrong type, but we can't solve that without adding generics
* The `$expr` query operator was marked as a `fieldQuery` when it is actually a `query` similar to `$and` and `$or`.

### DIFF
--- a/tests/Builder/Expression/MergeObjectsOperatorTest.php
+++ b/tests/Builder/Expression/MergeObjectsOperatorTest.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Tests\Builder\Expression;
+
+use MongoDB\Builder\Expression;
+use MongoDB\Builder\Pipeline;
+use MongoDB\Builder\Stage;
+use MongoDB\Tests\Builder\PipelineTestCase;
+
+/**
+ * Test $mergeObjects expression
+ */
+class MergeObjectsOperatorTest extends PipelineTestCase
+{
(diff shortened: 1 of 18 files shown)


GromNaN (approved):
It would be great to add tests on this operators to bury these changes.

alcaeus:
Fixed another issue with `$zip` (the `useLongestLength` argument should be optional) and added tests from the docs for all updated operators

tests/Builder/Expression/MergeObjectsOperatorTest.php
```diff
@@ -0,0 +1,39 @@
...
+ */
+class MergeObjectsOperatorTest extends PipelineTestCase
+{
+    public function testMergeObjects(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::lookup(
+                as: 'fromItems',
+                from: 'items',   // field in the orders collection
+                localField: 'item', // field in the items collection
```
GromNaN, line 23: You even took the comments 👏🏻

GromNaN (approved):

PR #21, opened 2024-01-11
PHPLIB-1334 Add tests on Logical Query Operators

GromNaN:
Fix [PHPLIB-1334](https://jira.mongodb.org/browse/PHPLIB-1334)

https://www.mongodb.com/docs/manual/reference/operator/query/#logical

- `$and` 
- `$not` (using the [syntax](https://www.mongodb.com/docs/manual/reference/operator/query/not/#mongodb-query-op.-not), as there is no example)
- `$nor`
- `$or`

### DIFF
--- a/tests/Builder/Query/Pipelines.php
+++ b/tests/Builder/Query/Pipelines.php
@@ -68,6 +68,82 @@ enum Pipelines: string
     JSON;
 
     /**
+     * AND Queries With Multiple Expressions Specifying the Same Field
+     *
+     * @see https://www.mongodb.com/docs/manual/reference/operator/query/and/#and-queries-with-multiple-expressions-specifying-the-same-field
+     */
+    case AndANDQueriesWithMultipleExpressionsSpecifyingTheSameField = <<<'JSON'
+    [
+        {
+            "$match": {
+                "$and": [
+                    {
+                        "price": {
+                            "$ne": {
+                                "$numberDouble": "1.9899999999999999911"
+                            }
+                        }
+                    },
+                    {
+                        "price": {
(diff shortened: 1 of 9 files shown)


tests/Builder/Query/Pipelines.php
```diff
@@ -67,6 +67,82 @@ enum Pipelines: string
...
+     */
+    case AndANDQueriesWithMultipleExpressionsSpecifyingTheSameField = <<<'JSON'
+    [
+        {
+            "$match": {
+                "$and": [
+                    {
+                        "price": {
+                            "$ne": {
+                                "$numberDouble": "1.9899999999999999911"
```
GromNaN, line 83: The yaml `1.99` is encoded as `{ "$numberDouble": "1.9899999999999999911" }`

```php
echo MongoDB\BSON\Document::fromPHP(['x' => 1.99])->toCanonicalExtendedJSON();

// { "x" : { "$numberDouble" : "1.9899999999999999911" } }
```

alcaeus: We could also deviate from the suggested value and choose a value that doesn't have this problem and can be represented exactly (e.g. `0.2`).

GromNaN: This is not an issue, as the value is always encoded the same way. But it can make tests to fail if we fix this behavior in a future version of the extension.

Tracking in [PHPC-2341](https://jira.mongodb.org/browse/PHPC-2341)
Other example: https://github.com/mongodb/mongo-php-builder/pull/22#discussion_r1450226432

alcaeus (approved):

PR #22, opened 2024-01-12
PHPLIB-1343 Add tests on Array Expression Operators

GromNaN:
Fix [PHPLIB-1343](https://jira.mongodb.org/browse/PHPLIB-1343)

- `$arrayElemAt`
- `$arrayToObject`
- `$concatArrays`
- `$filter`
- ~`$firstN`~ (already done)
- `$in`
- `$indexOfArray`
- `$isArray`
- ~`$lastN`~ (already done)
- `$map`
- `$maxN`
- `$minN`
- `$objectToArray` (skipped duplicate example with `$arrayToObject`)
- `$range`
- `$reduce`
- `$reverseArray`
- `$size`
- `$slice`
- `$sortArray`
- ~`$zip`~ (already done)

### DIFF
--- a/generator/config/expression/isArray.yaml
+++ b/generator/config/expression/isArray.yaml
@@ -3,7 +3,7 @@ name: $isArray
 link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/isArray/'
 type:
     - resolvesToBool
-encode: array
+encode: single
 description: |
     Determines if the operand is an array. Returns a boolean.
 arguments:
@@ -11,4 +11,23 @@ arguments:
         name: expression
         type:
             - expression
-        variadic: array
+tests:
+    -
+        name: 'Example'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/isArray/#example'
+        pipeline:
+            -
+                $project:
+                    items:
+                        $cond:
+                            if:
+                                $and:
+                                    -
+                                        $isArray: '$instock'
+                                    -
+                                        $isArray: '$ordered'
+                            then:
+                                $concatArrays:
+                                    - '$instock'
+                                    - '$ordered'
+                            else: 'One or more fields is not an array.'
--- a/generator/config/expression/isArray.yaml
+++ b/generator/config/expression/isArray.yaml
@@ -3,7 +3,7 @@ name: $isArray
 link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/isArray/'
 type:
     - resolvesToBool
-encode: single
+encode: array
 description: |
     Determines if the operand is an array. Returns a boolean.
 arguments:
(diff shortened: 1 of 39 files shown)


generator/config/expression/isArray.yaml
```diff
@@ -3,12 +3,31 @@ name: $isArray
 link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/isArray/'
 type:
     - resolvesToBool
-encode: array
+encode: single
```
GromNaN: The syntax [in the doc](https://www.mongodb.com/docs/manual/reference/operator/aggregation/isArray/) is confusing. It looks like the expression must be encapsulated in a single-value array, but in the example, the variable is used directly.
Maybe a doc issue.

tests/Builder/Expression/Pipelines.php
```diff
@@ -171,41 +490,705 @@ enum Pipelines: string
...
+                    "$map": {
+                        "input": "$tempsC",
+                        "as": "tempInCelsius",
+                        "in": {
+                            "$add": [
+                                {
+                                    "$multiply": [
+                                        "$$tempInCelsius",
+                                        {
+                                            "$numberDouble": "1.8000000000000000444"
```
GromNaN, line 569: Other example of numeric value incorrectly encoded.
Related to https://github.com/mongodb/mongo-php-builder/pull/21#discussion_r1448851990

src/Builder/BuilderEncoder.php
```diff
@@ -278,6 +278,8 @@ private function recursiveEncode(mixed $value): mixed
             foreach (get_object_vars($value) as $key => $val) {
                 $value->{$key} = $this->recursiveEncode($val);
             }
+
+            return $value;
```
GromNaN, line 282: Not a bug, but an optimization.

generator/config/expression/isArray.yaml
alcaeus: I believe this is covered by this note in the docs:
> Aggregation expressions accept a variable number of arguments. These arguments are normally passed as an array. However, when the argument is a single value, you can simplify your code by passing the argument directly without wrapping it in an array.

I'd have to do some tests on how values are handled. Looking at the examples in the [Behavior](https://www.mongodb.com/docs/manual/reference/operator/aggregation/isArray/#behavior) section, we should make sure that passing a value of `[1]` is actually recognised as an array, which might require changing `encode` back to `array` and updating the examples in the config files.

alcaeus (commented):
One suggestion regarding argument encoding for `$isArray`, but LGTM other than that.

generator/config/expression/isArray.yaml
GromNaN: Ok, reverted to always use the array syntax.

alcaeus (approved):

PR #23, opened 2024-01-15
PHPLIB-1335 Add tests on Element Query Operators

GromNaN:
Fix [PHPLIB-1335](https://jira.mongodb.org/browse/PHPLIB-1335)

https://www.mongodb.com/docs/manual/reference/operator/query/#element

- `$exists` (skipped the last redundant test)
- `$type`

### DIFF
--- a/tests/Builder/Query/TypeOperatorTest.php
+++ b/tests/Builder/Query/TypeOperatorTest.php
@@ -66,10 +66,10 @@ class TypeOperatorTest extends PipelineTestCase
     {
         $pipeline = new Pipeline(
             Stage::match(
-                zipCode: Query::type([2, 1]),
+                zipCode: Query::type(2, 1),
             ),
             Stage::match(
-                zipCode: Query::type(['string', 'double']),
+                zipCode: Query::type('string', 'double'),
             ),
         );
 
--- a/tests/Builder/Query/TypeOperatorTest.php
+++ b/tests/Builder/Query/TypeOperatorTest.php
@@ -1,0 +1,78 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Tests\Builder\Query;
+
+use MongoDB\Builder\Pipeline;
+use MongoDB\Builder\Query;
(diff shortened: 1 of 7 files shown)


tests/Builder/Query/TypeOperatorTest.php
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Test $type query
+ */
+class TypeOperatorTest extends PipelineTestCase
+{
+    public function testQueryingByArrayType(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::match(
+                zipCode: Query::type('array'),
```
alcaeus, line 21: As an idea for the future: we could provide an enum for the various BSON types supported by this operator here.

tests/Builder/Query/TypeOperatorTest.php
```diff
@@ -0,0 +1,78 @@
...
+        );
+
+        $this->assertSamePipeline(Pipelines::TypeQueryingByMinKeyAndMaxKey, $pipeline);
+    }
+
+    public function testQueryingByMultipleDataType(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::match(
+                zipCode: Query::type([2, 1]),
```
alcaeus: Would it make sense to make the type argument variadic?
```suggestion
                zipCode: Query::type(2, 1),
```

alcaeus (approved):
A couple of suggested improvements for the `$type` operator, but I think we can defer those to a separate ticket.

tests/Builder/Query/TypeOperatorTest.php
GromNaN: Ticket created: https://jira.mongodb.org/browse/PHPLIB-1363

tests/Builder/Query/TypeOperatorTest.php
GromNaN: Updated. The array for will always be used, even for a single value.

alcaeus (approved):

PR #24, opened 2024-01-15
PHPLIB-1336 Add tests on Evaluation Query Operators

GromNaN:
Fix [PHPLIB-1336](https://jira.mongodb.org/browse/PHPLIB-1336)

https://www.mongodb.com/docs/manual/reference/operator/query/#evaluation

- ~`$expr`~ (already done)
- `$jsonSchema`
- `$mod` contains some [error examples](https://www.mongodb.com/docs/manual/reference/operator/query/mod/#not-enough-elements-error) that are not allowed by the function signature.
- ~`$regex`~ (already done)
- `$text` requires a specific encoding, to prefix properties with dollar sign (`$`)
- `$where` the alternative example using `$expr` and `$function` is also implemented

### DIFF
--- a/generator/config/query/mod.yaml
+++ b/generator/config/query/mod.yaml
@@ -10,8 +10,33 @@ arguments:
     -
         name: divisor
         type:
-            - int
+            - number
     -
         name: remainder
         type:
-            - int
+            - number
+tests:
+    -
+        name: 'Use $mod to Select Documents'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/mod/#use--mod-to-select-documents'
+        pipeline:
+            -
+                $match:
+                    qty:
+                        $mod: [4, 0]
+    -
+        name: 'Floating Point Arguments'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/mod/#floating-point-arguments'
+        pipeline:
+            -
+                $match:
+                    qty:
+                        $mod: [4.0, 0]
+            -
+                $match:
+                    qty:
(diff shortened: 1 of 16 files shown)


generator/config/query/mod.yaml
```diff
@@ -10,8 +10,34 @@ arguments:
     -
         name: divisor
         type:
-            - int
+            - number
```
alcaeus, line 13: I suspected there were a few more that I missed previously. Thanks for finding those 🎉

tests/Builder/Query/WhereOperatorTest.php
```diff
@@ -0,0 +1,41 @@
...
+            ),
+            Stage::match(
+                Query::expr(
+                    Expression::function(
+                        body: <<<'JS'
+                            function(name) {
+                                return hex_md5(name) == "9b53e667f30cd329dca1ec9e6a83e994";
+                            }
+                            JS,
+                        args: ['$name'],
```
alcaeus: Should this be a fieldName expression?
```suggestion
                        args: [Expression::fieldName('name')],
```

GromNaN: This was `Expression::fieldPath('name')`

alcaeus (approved):

tests/Builder/Query/WhereOperatorTest.php
alcaeus: Yes, `fieldPath` was what I meant 🤦

PR #25, opened 2024-01-15
PHPLIB-1337 Add tests on Geospatial Query Operators

GromNaN:
Fix [PHPLIB-1337](https://jira.mongodb.org/browse/PHPLIB-1337)

https://www.mongodb.com/docs/manual/reference/operator/query/#geospatial

- `$geoIntersects`
- `$geoWithin`
- `$near` skipped "legacy coordinate", not supported
- `$nearSphere` skipped "legacy coordinate", not supported

### DIFF
--- a/generator/config/query/geoIntersects.yaml
+++ b/generator/config/query/geoIntersects.yaml
@@ -11,3 +11,42 @@ arguments:
         name: geometry
         type:
             - geometry
+tests:
+    -
+        name: 'Intersects a Polygon'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/geoIntersects/#intersects-a-polygon'
+        pipeline:
+            -
+                $match:
+                    loc:
+                        $geoIntersects:
+                            $geometry:
+                                type: 'Polygon'
+                                coordinates:
+                                    -
+                                        - [ 0, 0 ]
+                                        - [ 3, 6 ]
+                                        - [ 6, 1 ]
+                                        - [ 0, 0 ]
+    -
+        name: 'Intersects a Big Polygon'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/geoIntersects/#intersects-a--big--polygon'
+        pipeline:
+            -
+                $match:
+                    loc:
+                        $geoIntersects:
+                            $geometry:
+                                type: 'Polygon'
+                                coordinates:
+                                    -
+                                        - [ -100, 60 ]
+                                        - [ -100, 0 ]
+                                        - [ -100, -60 ]
+                                        - [ 100, -60 ]
+                                        - [ 100, 60 ]
+                                        - [ -100, 60 ]
+                                crs:
+                                    type: 'name'
+                                    properties:
+                                        name: 'urn:x-mongodb:crs:strictwinding:EPSG:4326'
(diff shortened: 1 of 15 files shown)


generator/config/query/near.yaml
```diff
@@ -14,14 +14,42 @@ arguments:
...
+        name: 'Query on Legacy Coordinates'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/near/#query-on-legacy-coordinates'
+        pipeline:
+            -
+                $match:
+                    location:
+                        $near:
+                            - -73.9667
+                            - 40.78
+                        $maxDistance: 0.1
```
GromNaN: This looks like a [bug in the example](https://www.mongodb.com/docs/manual/reference/operator/query/near/#query-on-legacy-coordinates), I need to check if the `$maxDistance` can be at the same level as `$near`.

generator/config/query/geoIntersects.yaml
```diff
@@ -11,3 +11,42 @@ arguments:
...
+tests:
+    -
+        name: 'Intersects a Polygon'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/geoIntersects/#intersects-a-polygon'
+        pipeline:
+            -
+                $match:
+                    loc:
+                        $geoIntersects:
+                            $geometry:
```
GromNaN, line 23: This requires the new `DollarObject` encoding type. https://github.com/mongodb/mongo-php-builder/pull/24/files#diff-bed00226517488779abea06643721208b8812c52dbf1c0d7f5aa01b87c5dfb69R114

GromNaN: No, it's actually documented. But I think we don't need to support eh "legacy coordinates"?

generator/config/query/near.yaml
GromNaN: No, it's actually documented. But I think we don't need to support the "legacy coordinates"?

src/Builder/BuilderEncoder.php
```diff
@@ -182,7 +183,19 @@ private function encodeAsDollarObject(OperatorInterface $value): stdClass
                 continue;
             }
 
-            $result->{'$' . $key} = $this->recursiveEncode($val);
+            $val = $this->recursiveEncode($val);
+
+            if ($key === 'geometry') {
```
alcaeus, line 188: We may want to consider introducing a special encoder for geo operators to extract this special logic.

alcaeus (approved):
LGTM - the special logic required for geometry keys makes a point for introducing a dedicated encoder down the line.

src/Builder/BuilderEncoder.php
GromNaN: For everything related to geometry, we need to be compatible with a GeoJSON library instead of creating a `$geometry` object.

alcaeus: I agree. Apart from a couple of libraries that saw their last updates 9 years ago, the only PHP libraries are found are [jmikola/geojson](https://github.com/jmikola/geojson) which is supported in Doctrine ODM, and [brick/geo](https://github.com/brick/geo) which is still in a pre-release (0.x) state but saw its last update a little more than a year ago. I'll defer to @jmikola's expertise on this.

GromNaN: Tracking in [PHPLIB-1364](https://jira.mongodb.org/browse/PHPLIB-1364)

PR #29, opened 2024-01-16
PHPLIB-1359 Add tests on Accumulators ($group, $bucket, $bucketAuto, $setWindowFields)

GromNaN:
Fix [PHPLIB-1359](https://jira.mongodb.org/browse/PHPLIB-1359)

https://www.mongodb.com/docs/manual/reference/operator/aggregation/#accumulators---group---bucket---bucketauto---setwindowfields-

For the operators that have accumulator and expression examples on the same page, I added the examples on both.

- ~`$accumulator`~ already done
- ~`$addToSet`~ already done
- `$avg` accumulator + expression
- `$bottom`
- `$bottomN`
- `$count` the empty object required modifications in Yaml parsing
- ~`$first`~ already done
- `$firstN` accumulator + expression
- ~`$last`~ already done
- `$lastN`
- `$max` accumulator + expression
- `$maxN`
- `$median` accumulator + expression
- `$mergeObjects`
- `$min` accumulator + expression
- `$percentile` accumulator + expression
- `$push`
- `$stdDevPop` accumulator + expression
- `$stdDevSamp` accumulator + expression
- `$sum` accumulator + expression
- `$top`
- `$topN`

### DIFF
--- a/generator/src/Definition/OperatorDefinition.php
+++ b/generator/src/Definition/OperatorDefinition.php
@@ -12,6 +12,7 @@ use function array_merge;
 use function array_values;
 use function assert;
 use function count;
+use function get_object_vars;
 use function sprintf;
 
 final class OperatorDefinition
@@ -47,7 +48,7 @@ final class OperatorDefinition
         // Optional arguments must be after required arguments
         $requiredArgs = $optionalArgs = [];
         foreach ($arguments as $arg) {
-            $arg = new ArgumentDefinition(...$arg);
+            $arg = new ArgumentDefinition(...get_object_vars($arg));
             if ($arg->optional) {
                 $optionalArgs[] = $arg;
             } else {
@@ -63,6 +64,9 @@ final class OperatorDefinition
 
         $this->arguments = array_merge($requiredArgs, $optionalArgs);
 
-        $this->tests = array_map(static fn (array $test): TestDefinition => new TestDefinition(...$test), array_values($tests));
+        $this->tests = array_map(
+            static fn (object $test): TestDefinition => new TestDefinition(...get_object_vars($test)),
+            array_values($tests),
+        );
     }
 }
--- a/generator/src/Definition/YamlReader.php
+++ b/generator/src/Definition/YamlReader.php
@@ -7,8 +7,7 @@ namespace MongoDB\CodeGenerator\Definition;
 use Symfony\Component\Finder\Finder;
 use Symfony\Component\Yaml\Yaml;
 
-use function assert;
-use function is_array;
+use function get_object_vars;
 
 final class YamlReader
 {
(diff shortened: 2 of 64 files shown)


generator/src/Definition/YamlReader.php
```diff
@@ -20,9 +17,8 @@ public function read(string $dirname): array
 
         $definitions = [];
         foreach ($finder as $file) {
-            $operator = Yaml::parseFile($file->getPathname());
-            assert(is_array($operator));
-            $definitions[] = new OperatorDefinition(...$operator);
+            $operator = Yaml::parseFile($file->getPathname(), Yaml::PARSE_OBJECT | Yaml::PARSE_OBJECT_FOR_MAP);
```
GromNaN, line 22: The option `PARSE_OBJECT_FOR_MAP` is required to ensure the `$count` empty object `{}` from the Yaml is not converted to an empty array. `[]`

generator/src/Definition/OperatorDefinition.php
```diff
@@ -63,6 +63,6 @@ public function __construct(
 
         $this->arguments = array_merge($requiredArgs, $optionalArgs);
 
-        $this->tests = array_map(static fn (array $test): TestDefinition => new TestDefinition(...$test), array_values($tests));
+        $this->tests = array_map(static fn (object $test): TestDefinition => new TestDefinition(...(array) $test), array_values($tests));
```
GromNaN: I think this line of code is "expert" level.

tests/Builder/Accumulator/AvgAccumulatorTest.php
```diff
@@ -0,0 +1,61 @@
...
+
+    public function testUseInSetWindowFieldsStage(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::setWindowFields(
+                partitionBy: Expression::fieldPath('state'),
+                sortBy: object(
+                    orderDate: 1,
+                ),
+                output: object(
```
GromNaN, line 48: I'm always using the same way to write the pipelines (using `object` and not `array` map), so I don't really cover all the feature.

tests/Builder/Expression/PercentileOperatorTest.php
```diff
@@ -0,0 +1,37 @@
...
+                _id: 0,
+                studentId: 1,
+                testPercentiles: Expression::percentile(
+                    input: [
+                        Expression::intFieldPath('test01'),
+                        Expression::longFieldPath('test02'),
+                        Expression::numberFieldPath('test03'),
+                    ],
+                    p: [0.5, 0.95],
+                    method: 'approximate',
```
GromNaN, line 30: This should be an enum.

generator/src/Definition/OperatorDefinition.php
alcaeus: Splitting this line across multiple lines may just make it a little less "expert" and help tell the closure apart from the arg:
```suggestion
        $this->tests = array_map(
            static fn (object $test): TestDefinition => new TestDefinition(...(array) $test),
            array_values($tests),
        );
```

alcaeus (approved):
Have one comment for legibility, but LGTM.

As for the topic of enums, I believe there are a couple of other places where we can consider using an enum (e.g. the `unbounded` and `current` identifiers for windows). These have the advantage of making possible values easier to discover.

GromNaN:
> As for the topic of enums, I believe there are a couple of other places where we can consider using an enum (e.g. the `unbounded` and `current` identifiers for windows). These have the advantage of making possible values easier to discover.

Since both values are often used together (at least in examples), we may also make an enum that return both.

PR #34, opened 2024-01-18
PHPLIB-1350 PHPLIB-1350 Add tests on Date Expression Operators and add Yaml tags for BSON types

GromNaN:
Fix [PHPLIB-1350](https://jira.mongodb.org/browse/PHPLIB-1350)

https://www.mongodb.com/docs/manual/reference/operator/aggregation/#date-expression-operators

- `$dateAdd`
- `$dateDiff`
- `$dateFromParts`
- `$dateFromString`
- `$dateSubtract`
- `$dateToParts`
- `$dateToString`
- `$dateTrunc`
- `$dayOfMonth` using verbose `date` field
- `$dayOfWeek` using verbose `date` field
- `$dayOfYear` using verbose `date` field
- `$hour` using verbose `date` field
- `$isoDayOfWeek` using verbose `date` field
- `$isoWeek` using verbose `date` field
- `$isoWeekYear` using verbose `date` field
- `$millisecond` using verbose `date` field
- `$minute` using verbose `date` field
- `$month` using verbose `date` field
- `$second` using verbose `date` field
- `$toDate`
- `$week` using verbose `date` field
- `$year` using verbose `date` field

### DIFF
--- a/generator/config/expression/dateAdd.yaml
+++ b/generator/config/expression/dateAdd.yaml
@@ -33,3 +33,100 @@ arguments:
         optional: true
         description: |
             The timezone to carry out the operation. $timezone must be a valid expression that resolves to a string formatted as either an Olson Timezone Identifier or a UTC Offset. If no timezone is provided, the result is displayed in UTC.
+tests:
+    -
+        name: 'Add a Future Date'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/dateAdd/#add-a-future-date'
+        pipeline:
+            -
+                $project:
+                    expectedDeliveryDate:
+                        $dateAdd:
+                            startDate: '$purchaseDate'
+                            unit: 'day'
+                            amount: 3
+            -
+                # $merge: 'shipping'
+                $merge:
+                    into: 'shipping'
+    -
+        name: 'Filter on a Date Range'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/dateAdd/#filter-on-a-date-range'
+        pipeline:
+            -
+                $match:
+                    $expr:
+                        $gt:
+                            - '$deliveryDate'
+                            -
+                                $dateAdd:
+                                    startDate: '$purchaseDate'
+                                    unit: 'day'
+                                    amount: 5
+            -
+                $project:
+                    _id: 0
+                    custId: 1
+                    purchased:
+                        $dateToString:
+                            format: '%Y-%m-%d'
+                            date: '$purchaseDate'
+                    delivery:
+                        $dateToString:
+                            format: '%Y-%m-%d'
+                            date: '$deliveryDate'
+    -
+        name: 'Adjust for Daylight Savings Time'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/dateAdd/#adjust-for-daylight-savings-time'
+        pipeline:
+            -
+                $project:
+                    _id: 0
+                    location: 1
+                    start:
+                        $dateToString:
+                            format: '%Y-%m-%d %H:%M'
+                            date: '$login'
+                    days:
+                        $dateToString:
+                            format: '%Y-%m-%d %H:%M'
+                            date:
+                                $dateAdd:
+                                    startDate: '$login'
+                                    unit: 'day'
+                                    amount: 1
+                                    timezone: '$location'
+                    hours:
+                        $dateToString:
+                            format: '%Y-%m-%d %H:%M'
+                            date:
+                                $dateAdd:
+                                    startDate: '$login'
+                                    unit: 'hour'
+                                    amount: 24
+                                    timezone: '$location'
+                    startTZInfo:
+                        $dateToString:
+                            format: '%Y-%m-%d %H:%M'
+                            date: '$login'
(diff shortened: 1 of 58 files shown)


generator/config/expression/dateFromString.yaml
```diff
@@ -44,3 +44,38 @@ arguments:
...
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/dateFromString/#onnull'
+        pipeline:
+            -
+                $project:
+                    date:
+                        $dateFromString:
+                            dateString: '$date'
+                            timezone: '$timezone'
+                            # onNull: new Date(0)
+                            onNull: 1970-01-01T00:00:00+00:00
```
GromNaN: I tried to represent the date using the Extended JSON format: `{ $date: 0 }`, but this get converted to `{ $date: { $numberInt: 0 } }` and the date is actually represented with a long `{ $date: { $numberLong: 0 } }`.

So I ended using Yaml date, that why I configured the parser to parse dates, and added a transformation pass using `array_walk_recursive` to convert `DateTimeInterface` into `MongoDB\BSON\UTCDateTime`.

generator/config/expression/dayOfMonth.yaml
```diff
@@ -22,3 +22,14 @@ arguments:
...
+    -
+        name: 'Example'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/dayOfMonth/#example'
+        pipeline:
+            -
+                $project:
+                    day:
+                        # $dayOfMonth: '$date'
+                        $dayOfMonth:
+                            date: '$date'
```
GromNaN, line 35: The [example](https://www.mongodb.com/docs/manual/reference/operator/aggregation/dayOfMonth/#example) is the same for all this date part operators. So I picked only the tested operator for each.

tests/Builder/Expression/DateAddOperatorTest.php
```diff
@@ -0,0 +1,124 @@
...
+ */
+class DateAddOperatorTest extends PipelineTestCase
+{
+    public function testAddAFutureDate(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::project(
+                expectedDeliveryDate: Expression::dateAdd(
+                    startDate: Expression::dateFieldPath('purchaseDate'),
+                    unit: 'day',
```
GromNaN, line 24: It would be very useful to have an enum for this units. [PHPLIB-1370](https://jira.mongodb.org/browse/PHPLIB-1370)

generator/config/expression/dateAdd.yaml
```diff
@@ -33,3 +33,100 @@ arguments:
...
+        pipeline:
+            -
+                $project:
+                    expectedDeliveryDate:
+                        $dateAdd:
+                            startDate: '$purchaseDate'
+                            unit: 'day'
+                            amount: 3
+            -
+                # $merge: 'shipping'
```
alcaeus, line 49: Noted you left a comment denoting the original since our `$merge` operator always uses the `into` argument.

generator/src/OperatorTestGenerator.php
```diff
@@ -83,7 +89,15 @@ public function createClass(GeneratorDefinition $definition, OperatorDefinition
...
             $caseName = str_replace([' ', '-'], '', ucwords(str_replace('$', '', $operator->name . ' ' . $test->name)));
 
-            $json = Document::fromPHP(['pipeline' => $test->pipeline])->toCanonicalExtendedJSON();
+            $pipeline = $this->convertTypeRecursively($test->pipeline);
+            $pipeline = $test->pipeline;
+            array_walk_recursive($pipeline, function (mixed &$value): void {
+                if ($value instanceof DateTimeInterface) {
+                    $value = new UTCDateTime($value);
+                }
+            });
```
alcaeus: This looks like it's left over from before extracting `convertTypeRecursively`.

generator/config/expression/dateFromString.yaml
alcaeus: Just adding a note that we can leverage a custom tag for this if we run into issues with the automatic parsing of dates when we don't expect it: https://symfony.com/doc/current/components/yaml.html#parsing-and-dumping-custom-tags

alcaeus (changes requested):

generator/src/OperatorTestGenerator.php
GromNaN: Removed, thanks.

generator/config/expression/dateFromString.yaml
GromNaN: I switched all specific BSON types to Yaml tags. That's more explicit.

GromNaN:
Added support for `!date 0` Yaml tag, and `!binary 'base64'` for https://github.com/mongodb/mongo-php-builder/pull/37#discussion_r1458006767

alcaeus (approved):
Yaml tags look a lot better, great work!

generator/config/expression/dateAdd.yaml
GromNaN: I did this in few other place. So that if we find there is a bug, it's easier to see the original example.

PR #35, opened 2024-01-18
PHPLIB-1348 Add tests on Custom Aggregation Expression Operators

GromNaN:
Fix [PHPLIB-1348](https://jira.mongodb.org/browse/PHPLIB-1348)

I updated the operators accepting `javascript` type, to automatically convert strings into a `MongoDB\BSON\Javascript` object.
The `$function` operator has default values for `args: []` and `lang: 'js'`, because this parameters are required. The lang is the only one supported.

https://www.mongodb.com/docs/manual/reference/operator/aggregation/#custom-aggregation-expression-operators

- `$accumulator` example updated to use automatic creation of `Javascript` objects from string
- `$function`
- `$where` fixed examples to be `Javascript` objects

### DIFF
--- a/generator/config/accumulator/accumulator.yaml
+++ b/generator/config/accumulator/accumulator.yaml
@@ -64,14 +64,29 @@ tests:
                     avgCopies:
                         $accumulator:
                             init:
-                                $code: 'function () { return { count: 0, sum: 0 } }'
+                                $code: |-
+                                    function() {
+                                        return { count: 0, sum: 0 }
+                                    }
                             accumulate:
-                                $code: 'function (state, numCopies) { return { count: state.count + 1, sum: state.sum + numCopies } }'
+                                $code: |-
+                                    function(state, numCopies) {
+                                        return { count: state.count + 1, sum: state.sum + numCopies }
+                                    }
                             accumulateArgs: [ "$copies" ],
                             merge:
-                                $code: 'function (state1, state2) { return { count: state1.count + state2.count, sum: state1.sum + state2.sum } }'
+                                $code: |-
+                                    function(state1, state2) {
+                                        return {
+                                            count: state1.count + state2.count,
+                                            sum: state1.sum + state2.sum
+                                        }
+                                    }
                             finalize:
-                                $code: 'function (state) { return (state.sum / state.count) }'
+                                $code: |-
+                                    function(state) {
+                                        return (state.sum / state.count)
+                                    }
                             lang: 'js'
 
     -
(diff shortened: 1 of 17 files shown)


src/Builder/Accumulator/AccumulatorAccumulator.php
```diff
@@ -70,20 +70,36 @@ public function __construct(
         Optional|PackedArray|ResolvesToArray|BSONArray|array $initArgs = Optional::Undefined,
         Optional|Javascript|string $finalize = Optional::Undefined,
     ) {
+        if (is_string($init)) {
+            $init = new Javascript($init);
+        }
+
         $this->init = $init;
+        if (is_string($accumulate)) {
+            $accumulate = new Javascript($accumulate);
```
GromNaN, line 80: The [`MongoDB\BSON\Javascript`](https://www.php.net/manual/en/class.mongodb-bson-javascript.php) constructor accepts a 2nd parameter `$scope`. I don't think we need to set a value.

jmikola: The scope has been deprecated since 4.4 ([SERVER-45460](https://jira.mongodb.org/browse/SERVER-45460)), so no need to support it.

generator/src/OperatorFactoryGenerator.php
```diff
@@ -80,6 +80,8 @@ private function addMethod(GeneratorDefinition $definition, OperatorDefinition $
             } else {
                 if ($argument->optional) {
                     $parameter->setDefaultValue(new Literal('Optional::Undefined'));
+                } elseif ($argument->default !== null) {
+                    $parameter->setDefaultValue($argument->default);
```
jmikola, line 84: Is this the first time this logic is required? The default value in the schema was previously `"string", "number", "boolean"` so I would have expected some handling for this already.

src/Builder/Accumulator/AccumulatorAccumulator.php
GromNaN: Thanks, that's what I just concluded by reading the [list of types](https://www.mongodb.com/docs/manual/reference/operator/query/type/#available-types).

generator/config/accumulator/accumulator.yaml
```diff
@@ -64,14 +64,29 @@ tests:
                     avgCopies:
                         $accumulator:
                             init:
-                                $code: 'function () { return { count: 0, sum: 0 } }'
+                                $code: |-
```
jmikola, line 67: Noted that this has no impact on `tests/Builder/Accumulator/Pipelines.php` apart from whitespace changes. Likewise for `query/where.yaml`.

jmikola (approved):
LGTM, but I don't understand the impact of the `OperatorFactoryGenerator.php` change.

generator/src/OperatorFactoryGenerator.php
GromNaN: This was a bug. The default value was set on the operator class contructor, but not on the factory method. There is duplicated code between both generators.

PR #37, opened 2024-01-18
PHPLIB-1339 Add tests on Bitwise Query Operators

GromNaN:
Fix [PHPLIB-1339](https://jira.mongodb.org/browse/PHPLIB-1339)

https://www.mongodb.com/docs/manual/reference/operator/query/#bitwise

- `$bitsAllClear`
- `$bitsAllSet`
- `$bitsAnyClear`
- `$bitsAnySet`

I made all this operators variadics, as [suggested previously for `$type`](https://github.com/mongodb/mongo-php-builder/pull/23#discussion_r1453187036).

### DIFF
--- a/generator/config/query/bitsAllClear.yaml
+++ b/generator/config/query/bitsAllClear.yaml
@@ -12,4 +12,29 @@ arguments:
         type:
             - int
             - binData
-            - array # of int
+        variadic: array
+tests:
+    -
+        name: 'Bit Position Array'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/bitsAllClear/#bit-position-array'
+        pipeline:
+            -
+                $match:
+                    a:
+                        $bitsAllClear: [1, 5]
+    -
+        name: 'Integer Bitmask'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/bitsAllClear/#integer-bitmask'
+        pipeline:
+            - $match:
+                    a:
+                        # $bitsAllClear: 35
+                        $bitsAllClear: [35]
+    -
+        name: 'BinData Bitmask'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/bitsAllClear/#bindata-bitmask'
+        pipeline:
+            - $match:
+                  a:
+                      $bitsAllClear:
+                          - { $binary: { base64: 'TUM9PQ==', subType: '00' } }
--- a/generator/config/query/bitsAllClear.yaml
+++ b/generator/config/query/bitsAllClear.yaml
@@ -37,4 +37,4 @@ tests:
             - $match:
                   a:
                       $bitsAllClear:
-                          - { $binary: { base64: 'TUM9PQ==', subType: '00' } }
+                          - !binary 'IA=='
--- a/generator/config/query/bitsAllClear.yaml
+++ b/generator/config/query/bitsAllClear.yaml
@@ -28,6 +28,7 @@ tests:
         pipeline:
             - $match:
                     a:
+                        # Example uses the short form, the builder always generates the array form
                         # $bitsAllClear: 35
                         $bitsAllClear: [35]
     -
--- a/generator/config/query/bitsAllClear.yaml
+++ b/generator/config/query/bitsAllClear.yaml
@@ -12,7 +12,7 @@ arguments:
         type:
             - int
             - binData
-        variadic: array
+            - array # of int
 tests:
     -
         name: 'Bit Position Array'
@@ -28,14 +28,11 @@ tests:
         pipeline:
             - $match:
                     a:
-                        # Example uses the short form, the builder always generates the array form
-                        # $bitsAllClear: 35
-                        $bitsAllClear: [35]
+                        $bitsAllClear: 35
     -
         name: 'BinData Bitmask'
         link: 'https://www.mongodb.com/docs/manual/reference/operator/query/bitsAllClear/#bindata-bitmask'
         pipeline:
             - $match:
                   a:
-                      $bitsAllClear:
-                          - !binary 'IA=='
+                      $bitsAllClear: !binary 'IA=='
--- a/generator/config/query/bitsAllSet.yaml
+++ b/generator/config/query/bitsAllSet.yaml
@@ -38,4 +38,4 @@ tests:
             - $match:
                   a:
                       $bitsAllSet:
-                          - !binary 'IA=='
+                          - !binary 'MA=='
--- a/generator/config/query/bitsAllSet.yaml
+++ b/generator/config/query/bitsAllSet.yaml
@@ -12,4 +12,29 @@ arguments:
         type:
             - int
             - binData
-            - array # of int
+        variadic: array
+tests:
+    -
+        name: 'Bit Position Array'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/bitsAllSet/#bit-position-array'
+        pipeline:
+            -
+                $match:
+                    a:
+                        $bitsAllSet: [1, 5]
+    -
+        name: 'Integer Bitmask'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/bitsAllSet/#integer-bitmask'
+        pipeline:
+            - $match:
+                  a:
+                      # $bitsAllSet: 50
+                      $bitsAllSet: [50]
+    -
+        name: 'BinData Bitmask'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/bitsAllSet/#bindata-bitmask'
+        pipeline:
+            - $match:
+                  a:
+                      $bitsAllSet:
+                          - { $binary: { base64: 'TUM9PQ==', subType: '00' } }
--- a/generator/config/query/bitsAllSet.yaml
+++ b/generator/config/query/bitsAllSet.yaml
@@ -12,7 +12,7 @@ arguments:
         type:
             - int
             - binData
-        variadic: array
+            - array # of int
 tests:
     -
         name: 'Bit Position Array'
@@ -28,14 +28,11 @@ tests:
         pipeline:
             - $match:
                   a:
-                      # Example uses the short form, the builder always generates the array form
-                      # $bitsAllSet: 50
-                      $bitsAllSet: [50]
+                      $bitsAllSet: 50
     -
         name: 'BinData Bitmask'
         link: 'https://www.mongodb.com/docs/manual/reference/operator/query/bitsAllSet/#bindata-bitmask'
         pipeline:
             - $match:
                   a:
-                      $bitsAllSet:
-                          - !binary 'MA=='
+                      $bitsAllSet: !binary 'MA=='
--- a/generator/config/query/bitsAllSet.yaml
+++ b/generator/config/query/bitsAllSet.yaml
@@ -28,6 +28,7 @@ tests:
         pipeline:
             - $match:
                   a:
+                      # Example uses the short form, the builder always generates the array form
                       # $bitsAllSet: 50
                       $bitsAllSet: [50]
     -
--- a/generator/config/query/bitsAllSet.yaml
+++ b/generator/config/query/bitsAllSet.yaml
@@ -37,4 +37,4 @@ tests:
             - $match:
                   a:
                       $bitsAllSet:
-                          - { $binary: { base64: 'TUM9PQ==', subType: '00' } }
+                          - !binary 'IA=='
--- a/generator/config/query/bitsAnyClear.yaml
+++ b/generator/config/query/bitsAnyClear.yaml
@@ -37,4 +37,4 @@ tests:
             - $match:
                   a:
                       $bitsAnyClear:
-                          - { $binary: { base64: 'TUM9PQ==', subType: '00' } }
+                          - !binary 'IA=='
--- a/generator/config/query/bitsAnyClear.yaml
+++ b/generator/config/query/bitsAnyClear.yaml
@@ -28,6 +28,7 @@ tests:
         pipeline:
             - $match:
                   a:
+                      # Example uses the short form, the builder always generates the array form
                       # $bitsAnyClear: 35
                       $bitsAnyClear: [35]
     -
--- a/generator/config/query/bitsAnyClear.yaml
+++ b/generator/config/query/bitsAnyClear.yaml
@@ -12,4 +12,29 @@ arguments:
         type:
             - int
             - binData
-            - array # of int
+        variadic: array
+tests:
+    -
+        name: 'Bit Position Array'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/bitsAnyClear/#bit-position-array'
+        pipeline:
+            -
+                $match:
+                    a:
+                        $bitsAnyClear: [1, 5]
+    -
+        name: 'Integer Bitmask'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/bitsAnyClear/#integer-bitmask'
+        pipeline:
+            - $match:
+                  a:
+                      # $bitsAnyClear: 35
+                      $bitsAnyClear: [35]
+    -
+        name: 'BinData Bitmask'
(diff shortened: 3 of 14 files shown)


generator/config/query/bitsAllClear.yaml
```diff
@@ -12,4 +12,29 @@ arguments:
...
+                        # $bitsAllClear: 35
+                        $bitsAllClear: [35]
+    -
+        name: 'BinData Bitmask'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/bitsAllClear/#bindata-bitmask'
+        pipeline:
+            - $match:
+                  a:
+                      $bitsAllClear:
+                          - { $binary: { base64: 'TUM9PQ==', subType: '00' } }
```
GromNaN: I've no idea what I'm doing with this binary data. This is the Extended JSON representation of the PHP `new Binary('MC==')`, which I want to be the equivalent of the JS `BinData(0, "ID==")`

jmikola: Note: [`Binary::__construct()`](https://www.php.net/manual/en/mongodb-bson-binary.construct.php) takes a string argument representing binary data. So if you're constructing this in PHP, you should be doing `new Binary(base64_decode(...))`.

I have no idea why the docs use base64 `ID==` as the representation of binary `00100000`. Even the shell seems to disagree:

```
> BinData(0, "ID==")
Binary.createFromBase64("IA==", 0)
```

Using a tool like [aaronluna.dev/base64](https://aaronluna.dev/base64), you can see that `ID==` has the following representation:

```
ASCII
HEX     2   0
8-BIT   00100000

6-BIT   001000 000011
DECIMAL 8      3
BASE64  I      D     =     =
```

`IA==` is:

```
ASCII
HEX     2   0
8-BIT   00100000

6-BIT   001000 000000
DECIMAL 8      0
BASE64  I      D     =     =
```

That suggests `ID==` is non-canonical, since it denotes 12 bits of data. Only by truncating the last four bits do you end up with the same value as `IA==`. I think this is worth a [DOCSP](https://jira.mongodb.org/browse/DOCSP) ticket to suggest a change to the server docs.

libbson actually raises an error from [`_bson_json_parse_binary_elem()`](https://github.com/mongodb/mongo-c-driver/blob/1.25.4/src/libbson/src/bson/bson-json.c#L849) when parsing `ID==` (perhaps rightfully so):

>  Invalid input string "ID==", looking for base64-encoded binary

I'd suggest switching over to `IA==` in this example and leaving a comment in the YAML explaining that you modified the value from the docs to use the more canonical representation.

generator/config/query/bitsAllSet.yaml
```diff
@@ -12,4 +12,29 @@ arguments:
...
+                      # $bitsAllSet: 50
+                      $bitsAllSet: [50]
+    -
+        name: 'BinData Bitmask'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/bitsAllSet/#bindata-bitmask'
+        pipeline:
+            - $match:
+                  a:
+                      $bitsAllSet:
+                          - { $binary: { base64: 'TUM9PQ==', subType: '00' } }
```
jmikola: https://www.mongodb.com/docs/manual/reference/operator/query/bitsAllSet/#bindata-bitmask uses `MC==`, which also isn't canonical:

```
> BinData(0, "MC==")
Binary.createFromBase64("MA==", 0)
```

I'd suggest including this in your DOCSP ticket as well. I've no idea where the author(s) got these values from.

generator/config/query/bitsAnyClear.yaml
```diff
@@ -12,4 +12,29 @@ arguments:
...
+                      # $bitsAnyClear: 35
+                      $bitsAnyClear: [35]
+    -
+        name: 'BinData Bitmask'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/bitsAnyClear/#bindata-bitmask'
+        pipeline:
+            - $match:
+                  a:
+                      $bitsAnyClear:
+                          - { $binary: { base64: 'TUM9PQ==', subType: '00' } }
```
jmikola: Same issue:

```
> BinData(0, "MC==")
Binary.createFromBase64("MA==", 0)
```

generator/config/query/bitsAnySet.yaml
```diff
@@ -12,4 +12,29 @@ arguments:
...
+                      # $bitsAnySet: 35
+                      $bitsAnySet: [35]
+    -
+        name: 'BinData Bitmask'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/bitsAnySet/#bindata-bitmask'
+        pipeline:
+            - $match:
+                  a:
+                      $bitsAnySet:
+                          - { $binary: { base64: 'TUM9PQ==', subType: '00' } }
```
jmikola: Same issue:

```
> BinData(0, "MC==")
Binary.createFromBase64("MA==", 0)
```

tests/Builder/Query/BitsAllClearOperatorTest.php
```diff
@@ -0,0 +1,52 @@
...
+ * Test $bitsAllClear query
+ */
+class BitsAllClearOperatorTest extends PipelineTestCase
+{
+    public function testBinDataBitmask(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::match(
+                a: Query::bitsAllClear(
+                    new Binary('MC=='),
```
jmikola: Per my earlier comment (https://github.com/mongodb/mongo-php-builder/pull/37/files#r1458006767), you'll need to use `base64_decode()` to prepare the binary string for the Binary constructor. Likewise for other test files.

And be sure to use the canonical base64 representations. I'm not sure how PHP handles this, but it'd be preferable not to rely on any behavior that truncates to 8-bit boundaries.

generator/config/query/bitsAllClear.yaml
GromNaN: Thanks, I'm fixing the BSON binary type in https://github.com/mongodb/mongo-php-builder/pull/34 using a Yaml tag.

The tag `!binary 'IA=='` will be converted to `new MongoDB\BSON\Binary(base64_decode('IA=='))`

GromNaN: See [DOCSP-35733](https://jira.mongodb.org/browse/DOCSP-35733)

generator/config/query/bitsAllClear.yaml
```diff
@@ -12,4 +12,29 @@ arguments:
...
+                        # $bitsAllClear: 35
+                        $bitsAllClear: [35]
+    -
+        name: 'BinData Bitmask'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/bitsAllClear/#bindata-bitmask'
+        pipeline:
+            - $match:
+                  a:
+                      $bitsAllClear:
+                          - !binary 'IA=='
```
jmikola: TIL about [YAML binary strings](https://yaml.org/type/binary.html#:~:text=Base64%20is%20the%20recommended%20way,e.g.%20the%20Java%20serialization%20format). Since this is a single line and we're not dealing with whitespace, I think it'd be reasonable to use the canonical `!!binary` notation.

But I'm also curious if a binary string is even appropriate here. The [corresponding example](https://www.mongodb.com/docs/manual/reference/operator/query/bitsAllClear/#bindata-bitmask) uses a BSON binary type, so shouldn't you be using extended JSON notation here? I think that was in an earlier draft of the PR.

Same question applies to other files.

generator/config/query/bitsAllClear.yaml
```diff
@@ -12,4 +12,29 @@ arguments:
...
+                    a:
+                        $bitsAllClear: [1, 5]
+    -
+        name: 'Integer Bitmask'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/bitsAllClear/#integer-bitmask'
+        pipeline:
+            - $match:
+                    a:
+                        # $bitsAllClear: 35
+                        $bitsAllClear: [35]
```
jmikola: What's the purpose of the array? The original [Integer Example](https://www.mongodb.com/docs/manual/reference/operator/query/bitsAllClear/#integer-bitmask) uses `35`, which seems to be what you have in the comment above.

Same question applies to other files.

generator/config/query/bitsAllSet.yaml
```diff
@@ -12,4 +12,29 @@ arguments:
...
+                      # $bitsAllSet: 50
+                      $bitsAllSet: [50]
+    -
+        name: 'BinData Bitmask'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/bitsAllSet/#bindata-bitmask'
+        pipeline:
+            - $match:
+                  a:
+                      $bitsAllSet:
+                          - !binary 'IA=='
```
jmikola: Note: apart from `bitsAllClear`, all of the other examples in the server docs used "MC==", which should be "MA==".

tests/Builder/Query/Pipelines.php
```diff
@@ -143,6 +143,282 @@ enum Pipelines: string
...
+     * @see https://www.mongodb.com/docs/manual/reference/operator/query/bitsAllClear/#bindata-bitmask
+     */
+    case BitsAllClearBinDataBitmask = <<<'JSON'
+    [
+        {
+            "$match": {
+                "a": {
+                    "$bitsAllClear": [
+                        {
+                            "$binary": {
```
jmikola: I see the `!binary` syntax in YAML _is_ producing extended JSON for a BSON binary type here. Where does that conversion happen? I'd have expected the YAML to just produce a binary string, which would still be parsed as a string type.

Do _any_ of the YAML templates include extended JSON?

Edit: I just saw https://github.com/mongodb/mongo-php-builder/pull/34/files#diff-a715bdcf88cd771d7ae09bd01863bab314ff194ba28e984689e9040d87846db9R20 and https://github.com/mongodb/mongo-php-builder/pull/34/files#diff-41189b0703bfb03f394530009a7d9c88c508f36a39585abd9f73017d6f77745eR135 from [PHPLIB-1350](https://jira.mongodb.org/browse/PHPLIB-1350) -- so this appears intentional.

tests/Builder/Query/BitsAllClearOperatorTest.php
```diff
@@ -0,0 +1,52 @@
...
+
+declare(strict_types=1);
+
+namespace MongoDB\Tests\Builder\Query;
+
+use MongoDB\BSON\Binary;
+use MongoDB\Builder\Pipeline;
+use MongoDB\Builder\Query;
+use MongoDB\Builder\Stage;
+use MongoDB\Tests\Builder\PipelineTestCase;
```
jmikola, line 11:
```suggestion
use MongoDB\Tests\Builder\PipelineTestCase;

use function base64_decode;
```

tests/Builder/Query/BitsAllSetOperatorTest.php
```diff
@@ -0,0 +1,52 @@
...
+
+declare(strict_types=1);
+
+namespace MongoDB\Tests\Builder\Query;
+
+use MongoDB\BSON\Binary;
+use MongoDB\Builder\Pipeline;
+use MongoDB\Builder\Query;
+use MongoDB\Builder\Stage;
+use MongoDB\Tests\Builder\PipelineTestCase;
```
jmikola, line 11:
```suggestion
use MongoDB\Tests\Builder\PipelineTestCase;

use function base64_decode;
```

tests/Builder/Query/BitsAnyClearOperatorTest.php
```diff
@@ -0,0 +1,52 @@
...
+
+declare(strict_types=1);
+
+namespace MongoDB\Tests\Builder\Query;
+
+use MongoDB\BSON\Binary;
+use MongoDB\Builder\Pipeline;
+use MongoDB\Builder\Query;
+use MongoDB\Builder\Stage;
+use MongoDB\Tests\Builder\PipelineTestCase;
```
jmikola, line 11:
```suggestion
use MongoDB\Tests\Builder\PipelineTestCase;

use function base64_decode;
```

tests/Builder/Query/BitsAllClearOperatorTest.php
```diff
@@ -0,0 +1,52 @@
...
+        );
+
+        $this->assertSamePipeline(Pipelines::BitsAllClearBitPositionArray, $pipeline);
+    }
+
+    public function testIntegerBitmask(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::match(
+                a: Query::bitsAllClear(35),
```
jmikola, line 48: Related to my earlier question about why the YAML pipeline wraps `35` in an array, is `Query::bitsAllClear()` actually yielding an array-wrapped value? If so, that seems inconsistent with what the server would expect: https://www.mongodb.com/docs/manual/reference/operator/query/bitsAllClear/#mongodb-query-op.-bitsAllClear

generator/config/query/bitsAllClear.yaml
GromNaN: Should I rename the `!binary` tag to `!bson_binary`, and the same for other tags?

generator/config/query/bitsAllClear.yaml
GromNaN: I added a comment. 

> Example uses the short form, the builder always generates the array form

generator/config/query/bitsAllSet.yaml
GromNaN: That's not very relevant for what we are testing, but I fixed. Thanks.

generator/config/query/bitsAllClear.yaml
jmikola: If these tags are custom on your part (wasn't clear to me if you were re-using existing YAML syntax), I think that'd be a good idea.

generator/config/query/bitsAllSet.yaml
jmikola: Aye, I don't imagine it affects the test, but thought it may be helpful to keep things in sync with the doc examples.

generator/config/query/bitsAllClear.yaml
jmikola: Apologies if I'm being obtuse, but `$bitsAllClear: [35]` seems entirely different than `$bitsAllClear: 35`. The former would be checking for the 35th bit and the latter checks for the bits according to the binary representation of `int(35)`. These are functionally different APIs (for all of these operators).

src/Builder/Query/BitsAllClearOperator.php
```diff
@@ -28,16 +25,21 @@ class BitsAllClearOperator implements FieldQueryInterface, OperatorInterface
...
+    /** @var list<Binary|int|non-empty-string> $bitmask */
+    public readonly array $bitmask;
 
     /**
-     * @param BSONArray|Binary|PackedArray|array|int|non-empty-string $bitmask
+     * @param Binary|int|non-empty-string ...$bitmask
+     * @no-named-arguments
      */
-    public function __construct(Binary|PackedArray|BSONArray|array|int|string $bitmask)
+    public function __construct(Binary|int|string ...$bitmask)
```
jmikola: Re:

 * https://github.com/mongodb/mongo-php-builder/pull/37/files#r1459574672
 * https://github.com/mongodb/mongo-php-builder/pull/37/files#r1459323536

I think the problem I'm pointing out in the above threads is due to ambiguity when a single integer is passed here. If `5` is passed to this method, it is interpreted as `[5]`, which is not the same.

I think this needs to be reworked to accept either:

 * An array of ints
 * A single int
 * A single BSON Binary

GromNaN: You're right, there are 3 distinct behaviors. I'll remove the variadic and change the type to `int|binary|list<int>`.

generator/config/query/bitsAllClear.yaml
GromNaN: Thanks for insisting, it's fixed.

generator/config/query/bitsAllClear.yaml
GromNaN: See https://github.com/mongodb/mongo-php-builder/pull/43

jmikola (approved):

PR #38, opened 2024-01-18
PHPLIB-1363 PHPLIB-1355 Add enum for `$type` query and `$meta` expression

GromNaN:
Fix [PHPLIB-1363](https://jira.mongodb.org/browse/PHPLIB-1363)
Fix [PHPLIB-1355](https://jira.mongodb.org/browse/PHPLIB-1355)

The enum implements `FieldQueryInterface` so it can be used anywhere a field query is expected. Deprecated types have been skipped.

List imported from [the doc](https://www.mongodb.com/docs/manual/reference/operator/query/type/#available-types), including the alias `number`.

Usage:
```php
        $pipeline = new Pipeline(
            Stage::match(
                zipCode: QueryType::Int,
            ),
        );
```

Limitation, it's not possible to use the enum in a list.
```php
        $pipeline = new Pipeline(
            Stage::match(
                zipCode: [QueryType::Int, QueryType::Long],
            ),
        );
```

### DIFF
--- a/src/Builder/Meta.php
+++ b/src/Builder/Meta.php
@@ -31,6 +31,11 @@ enum Meta: string implements OperatorInterface, ExpressionInterface
      */
     case IndexKey = 'indexKey';
 
+    /**
+     * Display search terms in their original context.
+     */
+    case SearchHighlights = 'searchHighlights';
+
     public function getOperator(): string
     {
         return '$meta';
--- a/src/Builder/Meta.php
+++ b/src/Builder/Meta.php
@@ -1,0 +1,38 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Builder;
+
+use MongoDB\Builder\Type\Encode;
+use MongoDB\Builder\Type\ExpressionInterface;
+use MongoDB\Builder\Type\OperatorInterface;
+
+/**
+ * Returns the metadata associated with a document, e.g. "textScore" when performing text search.
+ *
+ * @see https://www.mongodb.com/docs/manual/reference/operator/aggregation/meta/
+ */
+enum Meta: string implements OperatorInterface, ExpressionInterface
+{
+    public const ENCODE = Encode::Single;
+
+    /**
+     * Returns the score associated with the corresponding $text query for each
+     * matching document. The text score signifies how well the document matched
+     * the search term or terms.
+     */
+    case TextScore = 'textScore';
+
+    /**
+     * Returns an index key for the document if a non-text index is used. The
+     * { $meta: "indexKey" } expression is for debugging purposes only, and
+     * not for application logic, and is preferred over cursor.returnKey().
+     */
+    case IndexKey = 'indexKey';
+
+    public function getOperator(): string
+    {
+        return '$meta';
+    }
+}
--- a/src/Builder/QueryType.php
+++ b/src/Builder/QueryType.php
@@ -19,27 +19,25 @@ enum QueryType: string implements OperatorInterface, FieldQueryInterface
 {
     public const ENCODE = Encode::Array;
 
-    case Double = 'double';
-    case String = 'string';
-    case Object = 'object';
     case Array = 'array';
-    case BinaryData = 'binData';
-    case ObjectId = 'objectId';
-    case Boolean = 'bool';
-    case Date = 'date';
-    case Null = 'null';
-    case Regex = 'regex';
-    case JavaScript = 'javascript';
-    case Int = 'int';
-    case Timestamp = 'timestamp';
-    case Long = 'long';
+    case Binary = 'binData';
+    case Bool = 'bool';
+    case Decimal64 = 'double';
     case Decimal128 = 'decimal';
-    case MinKey = 'minKey';
+    case Int32 = 'int';
+    case Int64 = 'long';
+    case Javascript = 'javascript';
+    case Object = 'object';
+    case ObjectId = 'objectId';
     case MaxKey = 'maxKey';
-    /**
-     * Alias for Double, Int, Long, Decimal
-     */
+    case MinKey = 'minKey';
+    case Null = 'null';
+    /** Alias for Int32, Int64, Decimal64, Decimal128 */
     case Number = 'number';
+    case Regex = 'regex';
+    case String = 'string';
+    case Timestamp = 'timestamp';
+    case UTCDateTime = 'date';
 
     public function getOperator(): string
     {
--- a/src/Builder/QueryType.php
+++ b/src/Builder/QueryType.php
@@ -1,0 +1,48 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Builder;
+
+use MongoDB\Builder\Query\TypeOperator;
+use MongoDB\Builder\Type\Encode;
+use MongoDB\Builder\Type\FieldQueryInterface;
+use MongoDB\Builder\Type\OperatorInterface;
+
+/**
+ * Shortcut for $type field query operator
+ *
+ * @see https://www.mongodb.com/docs/manual/reference/operator/query/type/#available-types
+ * @see TypeOperator
+ */
+enum QueryType: string implements OperatorInterface, FieldQueryInterface
+{
+    public const ENCODE = Encode::Array;
+
+    case Double = 'double';
+    case String = 'string';
+    case Object = 'object';
+    case Array = 'array';
+    case BinaryData = 'binData';
+    case ObjectId = 'objectId';
+    case Boolean = 'bool';
+    case Date = 'date';
+    case Null = 'null';
+    case Regex = 'regex';
+    case JavaScript = 'javascript';
+    case Int = 'int';
+    case Timestamp = 'timestamp';
+    case Long = 'long';
+    case Decimal128 = 'decimal';
+    case MinKey = 'minKey';
+    case MaxKey = 'maxKey';
+    /**
+     * Alias for Double, Int, Long, Decimal
+     */
+    case Number = 'number';
+
+    public function getOperator(): string
+    {
+        return '$type';
+    }
+}
--- a/tests/Builder/Query/TypeOperatorTest.php
+++ b/tests/Builder/Query/TypeOperatorTest.php
@@ -6,6 +6,7 @@ namespace MongoDB\Tests\Builder\Query;
 
 use MongoDB\Builder\Pipeline;
 use MongoDB\Builder\Query;
+use MongoDB\Builder\QueryType;
 use MongoDB\Builder\Stage;
 use MongoDB\Tests\Builder\PipelineTestCase;
 
@@ -32,16 +33,16 @@ class TypeOperatorTest extends PipelineTestCase
                 zipCode: Query::type(2),
             ),
             Stage::match(
-                zipCode: Query::type('string'),
+                zipCode: QueryType::String,
             ),
             Stage::match(
                 zipCode: Query::type(1),
             ),
             Stage::match(
-                zipCode: Query::type('double'),
+                zipCode: Query::type(QueryType::Double),
             ),
             Stage::match(
-                zipCode: Query::type('number'),
+                zipCode: QueryType::Number,
             ),
         );
 
@@ -69,7 +70,7 @@ class TypeOperatorTest extends PipelineTestCase
                 zipCode: Query::type(2, 1),
             ),
             Stage::match(
-                zipCode: Query::type('string', 'double'),
+                zipCode: Query::type(QueryType::String, 'double'),
             ),
         );
 
--- a/tests/Builder/Query/TypeOperatorTest.php
+++ b/tests/Builder/Query/TypeOperatorTest.php
@@ -39,7 +39,7 @@ class TypeOperatorTest extends PipelineTestCase
                 zipCode: Query::type(1),
             ),
             Stage::match(
-                zipCode: Query::type(QueryType::Double),
+                zipCode: Query::type(QueryType::Decimal64),
             ),
             Stage::match(
                 zipCode: QueryType::Number,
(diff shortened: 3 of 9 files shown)


src/Builder/BuilderEncoder.php
```diff
@@ -101,6 +102,10 @@ public function encode($value): stdClass|array|string
             throw new LogicException(sprintf('Class "%s" does not implement OperatorInterface.', $value::class));
         }
 
+        if ($value instanceof BackedEnum) {
+            return $this->encodeAsEnum($value);
```
jmikola: The naming of this method seems odd to me, as the other `encodeAs` methods are consistent with `$value::ENCODE`. Would `encodeEnum()` make more sense here? Alternatively, I think the method might not even be required at all.

I was a bit confused by the difference in how the `Encode::Array` case is handled for a BackedEnum in `encodeAsEnum()` vs. other types in the `switch` below. Looking at `encodeAsArray()`, it creates a list of the object's properties. In `encodeAsEnum()`, you explicitly extract the `value` property (presumably because the `name` property is irrelevant) and then apply the handling we see in `encodeAsArray()` or `encodeAsSingle()`.

Would it make more sense to eagerly convert the BackedEnum into an object with a single `value` property, and then feed that to the switch statement below?

I imagine a BackedEnum has no business being used with other encoding types, such as `Encode::Group`, but I'm not sure if you'd want to assert on that or just trust the `switch` logic to produce something that server would ultimately reject.

src/Builder/QueryType.php
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Shortcut for $type field query operator
+ *
+ * @see https://www.mongodb.com/docs/manual/reference/operator/query/type/#available-types
+ * @see TypeOperator
+ */
+enum QueryType: string implements OperatorInterface, FieldQueryInterface
+{
+    public const ENCODE = Encode::Array;
```
jmikola, line 20: Is this the first time a BackedEnum is being used for an encoding type? If so, perhaps it'd be prudent to have `BuilderEncoder::encode()` assert that that `$value::ENCODE` is `Encode::Array` (and you can revise that as needed should another use case come up).

That would address my question about how a BackedEnum might be handled for other `switch` cases.

src/Builder/BuilderEncoder.php
GromNaN: Thanks for the review. I found a simpler implementation by only checking if `$value` is a `BackedEnum` in `encodeAsArray`. I don't need to implement other encoding methods for now, as it's the only enum I have.

tests/Builder/Query/TypeOperatorTest.php
```diff
@@ -32,7 +33,7 @@ public function testQueryingByDataType(): void
                 zipCode: Query::type(2),
             ),
             Stage::match(
-                zipCode: Query::type('string'),
+                zipCode: QueryType::String,
```
alcaeus, line 36: Huh, this is interesting. I had initially considered using an enum case as the argument for `Query::type`, but this is interesting as well.

GromNaN: I can customize the `Query::type` factory to accept enum values and convert them to string if necessary.

alcaeus: Not sure if that is necessary. This allows people to do whatever they want with the `$type` operator, but provides a nice shorthand for all known types.

src/Builder/BuilderEncoder.php
```diff
@@ -129,6 +130,10 @@ public function encode($value): stdClass|array|string
      */
     private function encodeAsArray(OperatorInterface $value): stdClass
     {
+        if ($value instanceof BackedEnum) {
+            return $this->wrap($value, [$value->value]);
```
jmikola: Is the NoInterfaceProperties error below the result of Psalm failing to infer that BackedEnum has an implicit `$value` property?

src/Builder/BuilderEncoder.php
```diff
@@ -206,6 +211,10 @@ private function encodeAsDollarObject(OperatorInterface $value): stdClass
      */
     private function encodeAsSingle(OperatorInterface $value): stdClass
     {
+        if ($value instanceof BackedEnum) {
+            return $this->wrap($value, $value->value);
```
jmikola: Is this code path used? Your earlier [comment](https://github.com/mongodb/mongo-php-builder/pull/38#discussion_r1457938643) suggested that only `encodeAsArray()` would need to check for a BackedEnum instance.

tests/Builder/Query/TypeOperatorTest.php
jmikola: I assume this only works because QueryType is both an enum _and_ an OperatorInterface (and FieldQueryInterface in that it's associated with a field path). I assume this convenience only works for a single type check, though. Trying to match `zipCode` with `[QueryType::String, ...]` in order to match multiple types would produce invalid syntax, correct?

And that's where `Query::type(...)` comes in.

jmikola (commented):
Per the Psalm error below, what enforces that OperatorInterface instances define an `ENCODE` constant?

LGTM, but I think the various Psalm errors need correction.

src/Builder/BuilderEncoder.php
GromNaN: Indeed it was implemented because that was trivial. I added an enum for `$meta`, that uses this single-value encoder.

src/Builder/BuilderEncoder.php
GromNaN: Yes it looks like an issue in palm: https://github.com/vimeo/psalm/issues/10585

phpstan don't have this issue: https://phpstan.org/r/e7094884-1adf-4369-b255-6c7e45367865

src/Builder/QueryType.php
GromNaN: We need to rework the whole encoder class. The "ENCODE" constant will not stay, it may become a method of the Operator interface.

src/Builder/BuilderEncoder.php
jmikola: How is this related to the NoInterfaceProperties warning earlier in `encode()`? There, you're only checking for an instance of `FieldPathInterface` and then accessing a `$value` property. That seems incorrect for an entirely different reason.

src/Builder/QueryType.php
jmikola: If this isn't already ticketed please do so and drop a reference to the issue.

src/Builder/Meta.php
```diff
@@ -0,0 +1,38 @@
...
+use MongoDB\Builder\Type\Encode;
+use MongoDB\Builder\Type\ExpressionInterface;
+use MongoDB\Builder\Type\OperatorInterface;
+
+/**
+ * Returns the metadata associated with a document, e.g. "textScore" when performing text search.
+ *
+ * @see https://www.mongodb.com/docs/manual/reference/operator/aggregation/meta/
+ */
+enum Meta: string implements OperatorInterface, ExpressionInterface
```
GromNaN, line 16: @alcaeus suggested to name it `SortMeta`. https://github.com/mongodb/mongo-php-builder/pull/56#discussion_r1467476295

jmikola: [$meta: Behavior](https://www.mongodb.com/docs/manual/reference/operator/aggregation/meta/#behavior) suggests the operator can be used in context beyond sorting, such as projections.

src/Builder/Meta.php
```diff
@@ -0,0 +1,38 @@
...
+     * the search term or terms.
+     */
+    case TextScore = 'textScore';
+
+    /**
+     * Returns an index key for the document if a non-text index is used. The
+     * { $meta: "indexKey" } expression is for debugging purposes only, and
+     * not for application logic, and is preferred over cursor.returnKey().
+     */
+    case IndexKey = 'indexKey';
```
jmikola, line 32: Do the cases here preclude users from using this with additional modes? [$meta: Definition](https://www.mongodb.com/docs/manual/reference/operator/aggregation/meta/#definition) mentions [searchScore](https://www.mongodb.com/docs/atlas/reference/full-text-search/scoring/) and [searchHighlights](https://www.mongodb.com/docs/atlas/reference/full-text-search/highlighting/) for Atlas.

jmikola (commented):
Just realized I had some pending comments that were never posted. Doing so now.

GromNaN:
Rebased on top of https://github.com/mongodb/mongo-php-builder/pull/46

src/Builder/QueryType.php
```diff
@@ -0,0 +1,48 @@
...
+    case Regex = 'regex';
+    case JavaScript = 'javascript';
+    case Int = 'int';
+    case Timestamp = 'timestamp';
+    case Long = 'long';
+    case Decimal128 = 'decimal';
+    case MinKey = 'minKey';
+    case MaxKey = 'maxKey';
+    /**
+     * Alias for Double, Int, Long, Decimal
```
jmikola:
```suggestion
     * Alias for Double, Int, Long, Decimal128
```

I'm not sure if you intended for this to refer to the other cases, or the backed values. If the backed values, then I'd suggest not capitalizing these at all.

src/Builder/QueryType.php
```diff
@@ -0,0 +1,48 @@
...
+ */
+enum QueryType: string implements OperatorInterface, FieldQueryInterface
+{
+    public const ENCODE = Encode::Array;
+
+    case Double = 'double';
+    case String = 'string';
+    case Object = 'object';
+    case Array = 'array';
+    case BinaryData = 'binData';
```
jmikola:
```suggestion
    case Binary = 'binData';
```

Is "Data" redundant here? This would also be consistent with our BSON class name.

src/Builder/QueryType.php
```diff
@@ -0,0 +1,48 @@
...
+    case ObjectId = 'objectId';
+    case Boolean = 'bool';
+    case Date = 'date';
+    case Null = 'null';
+    case Regex = 'regex';
+    case JavaScript = 'javascript';
+    case Int = 'int';
```
jmikola:
```suggestion
    case Int = 'int';
```

Would Int32 be more clear to users?

src/Builder/QueryType.php
```diff
@@ -0,0 +1,48 @@
...
+    case Timestamp = 'timestamp';
+    case Long = 'long';
```
jmikola:
```suggestion
    case Int64 = 'long';
```

This also leaves nothing to the imagination.

src/Builder/QueryType.php
```diff
@@ -0,0 +1,48 @@
...
+    case String = 'string';
+    case Object = 'object';
+    case Array = 'array';
+    case BinaryData = 'binData';
+    case ObjectId = 'objectId';
+    case Boolean = 'bool';
+    case Date = 'date';
+    case Null = 'null';
+    case Regex = 'regex';
+    case JavaScript = 'javascript';
```
jmikola:
```suggestion
    case Javascript = 'javascript';
```

Consistent with out BSON class name. Feel free to disregard if you prefer the existing casing, though.

src/Builder/QueryType.php
```diff
@@ -0,0 +1,48 @@
...
+    public const ENCODE = Encode::Array;
+
+    case Double = 'double';
+    case String = 'string';
+    case Object = 'object';
+    case Array = 'array';
+    case BinaryData = 'binData';
+    case ObjectId = 'objectId';
+    case Boolean = 'bool';
+    case Date = 'date';
```
jmikola:
```suggestion
    case UTCDateTime = 'date';
```

If you want to be consistent with our BSON class. Feel free to disregard.

src/Builder/QueryType.php
```diff
@@ -0,0 +1,48 @@
...
+    case Null = 'null';
+    case Regex = 'regex';
+    case JavaScript = 'javascript';
+    case Int = 'int';
+    case Timestamp = 'timestamp';
+    case Long = 'long';
+    case Decimal128 = 'decimal';
```
jmikola: I do think using the BSON class name was a good idea here, as users could confuse "decimal" with "double" if they didn't realize there was a specific type.

src/Builder/QueryType.php
```diff
@@ -0,0 +1,48 @@
...
+{
+    public const ENCODE = Encode::Array;
+
+    case Double = 'double';
+    case String = 'string';
+    case Object = 'object';
+    case Array = 'array';
+    case BinaryData = 'binData';
+    case ObjectId = 'objectId';
+    case Boolean = 'bool';
```
jmikola:
```suggestion
    case Bool = 'bool';
```

I don't feel strongly about this. We don't have a BSON class for this, but `bool` seems to be the canonical name for PHP typing.

jmikola (commented):
Left some additional comments on the enum cases, and there are already a number of open comment threads from last week.

src/Builder/QueryType.php
jmikola: It just occurred to me that you might have chosen these names to be consistent with the ExpressionFactoryTrait field path methods. Unless it was just the first column from the table in  [$type: Available Types](https://www.mongodb.com/docs/manual/reference/operator/query/type/#available-types).

In any event, you may want to keep the factory methods consistent with the enums here. And if factory methods will be distinguishing 32-bit and 64-bit integers, I do think explicit names with the bit size are more clear to users than "int" and "long".

src/Builder/Meta.php
GromNaN: The `ExpressionInterface` allows to use it anywhere an expression is expected.

src/Builder/QueryType.php
GromNaN: I updated all the type name to take more descriptive ones, from the BSON classes when they exist.

GromNaN:
Closing for now. Lets revisit later. Ticket still open.

PR #42, opened 2024-01-19
PHPLIB-1354 Add tests on String Expression Operators

GromNaN:
Fix [PHPLIB-1354](https://jira.mongodb.org/browse/PHPLIB-1354)

https://www.mongodb.com/docs/manual/reference/operator/aggregation/#string-expression-operators

- `$concat`
- ~`$dateFromString`~ already done
- ~`$dateToString`~ already done
- `$indexOfBytes`
- `$indexOfCP`
- `$ltrim`
- `$regexFind` Skipped repetitive tests on regex flags, but implemented all the ways to set the regex.
- `$regexFindAll` Skipped repetitive tests on regex flags, but implemented all the ways to set the regex.
- `$regexMatch` Skipped repetitive tests on regex flags, but implemented all the ways to set the regex.
- `$replaceOne` Skipped repetitive tests on regex flags, but implemented all the ways to set the regex.
- `$replaceAll`
- `$rtrim`
- `$split`
- `$strLenBytes`
- `$strLenCP`
- `$strcasecmp`
- `$substr`
- `$substrBytes`
- `$substrCP`
- `$toLower`
- `$toString`
- `$trim`
- `$toUpper`

Skipped repetitive tests on regex flags, but implemented all the ways to set the regex.

### DIFF
--- a/generator/config/expression/concat.yaml
+++ b/generator/config/expression/concat.yaml
@@ -12,3 +12,15 @@ arguments:
         type:
             - resolvesToString
         variadic: array
+tests:
+    -
+        name: 'Example'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/concat/#examples'
+        pipeline:
+            -
+                $project:
+                    itemDescription:
+                        $concat:
+                            - '$item'
+                            - ' - '
+                            - '$description'
--- a/generator/config/expression/regexFindAll.yaml
+++ b/generator/config/expression/regexFindAll.yaml
@@ -44,7 +44,7 @@ tests:
             -
                 $addFields:
                     returnObject:
-                        # Specify i as part of the regex field
+                        # Specify i as part of the regex type
                         $regexFindAll:
                             input: '$description'
                             regex: !regex ['line', 'i']
@@ -59,7 +59,7 @@ tests:
             -
                 $addFields:
                     returnObject:
-                        # Specify i in the options field
+                        # Mix Regex type with options field
                         $regexFindAll:
                             input: '$description'
                             regex: !regex 'line'
--- a/generator/config/expression/regexFindAll.yaml
+++ b/generator/config/expression/regexFindAll.yaml
@@ -26,3 +26,74 @@ arguments:
         type:
             - string
         optional: true
+tests:
+    -
+        name: '$regexFindAll and Its Options'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/regexFindAll/#-regexfindall-and-its-options'
+        pipeline:
+            -
+                $addFields:
+                    returnObject:
+                        $regexFindAll:
+                            input: '$description'
+                            regex: !regex 'line'
+    -
+        name: 'i Option'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/regexFindAll/#i-option'
+        pipeline:
+            -
+                $addFields:
+                    returnObject:
+                        # Specify i as part of the regex field
+                        $regexFindAll:
+                            input: '$description'
+                            regex: !regex ['line', 'i']
+            -
+                $addFields:
+                    returnObject:
+                        # Specify i in the options field
+                        $regexFindAll:
+                            input: '$description'
+                            regex: 'line'
+                            options: 'i'
+            -
+                $addFields:
+                    returnObject:
+                        # Specify i in the options field
+                        $regexFindAll:
+                            input: '$description'
+                            regex: !regex 'line'
+                            options: 'i'
+    -
+        name: 'Use $regexFindAll to Parse Email from String'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/regexFindAll/#use--regexfindall-to-parse-email-from-string'
+        pipeline:
+            -
+                $addFields:
+                    email:
+                        $regexFindAll:
+                            input: '$comment'
+                            regex: !regex ['[a-z0-9_.+-]+@[a-z0-9_.+-]+\.[a-z0-9_.+-]+', 'i']
+            -
+                $set:
+                    email: '$email.match'
+    -
+        name: 'Use Captured Groupings to Parse User Name'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/regexFindAll/#use-captured-groupings-to-parse-user-name'
+        pipeline:
+            -
+                $addFields:
+                    names:
+                        $regexFindAll:
+                            input: '$comment'
+                            regex: !regex ['([a-z0-9_.+-]+)@[a-z0-9_.+-]+\.[a-z0-9_.+-]+', 'i']
+            -
+                $set:
+                    names:
+                        $reduce:
+                            input: '$names.captures'
+                            initialValue: []
+                            in:
+                                $concatArrays:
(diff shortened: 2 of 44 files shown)


generator/config/expression/concat.yaml
```diff
@@ -12,3 +12,15 @@ arguments:
         type:
             - resolvesToString
         variadic: array
+tests:
+    -
+        name: 'Example'
```
jmikola, line 17: Remind me: these names are just used to produce the constants in the generated pipeline file, correct? You apply `ucfirst()` to the base filename and then append the test name for each pipeline to generate the constant identifier.

generator/config/expression/regexFindAll.yaml
```diff
@@ -26,3 +26,74 @@ arguments:
...
+                    returnObject:
+                        # Specify i in the options field
+                        $regexFindAll:
+                            input: '$description'
+                            regex: 'line'
+                            options: 'i'
+            -
+                $addFields:
+                    returnObject:
+                        # Specify i in the options field
```
jmikola: This comment can be distinguished from the previous in that you're specifying the option alongside an actual regex BSON type. Likewise, maybe the first example should specify "part of a BSON regex type" instead of "part of a regex field".

Also applies to other files below.

generator/config/expression/substrBytes.yaml
```diff
@@ -23,3 +23,25 @@ arguments:
             - resolvesToInt
         description: |
             If length is a negative number, $substr returns a substring that starts at the specified index and includes the rest of the string.
+tests:
+    -
+        name: 'Single-Byte Character Set'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/substrBytes/#single-byte-character-set'
```
jmikola, line 29:
```suggestion
        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/substrBytes/#example'
```

generator/config/expression/substrBytes.yaml
```diff
@@ -23,3 +23,25 @@ arguments:
...
+                            - 2
+                    quarterSubtring:
+                        $substrBytes:
+                            - '$quarter'
+                            - 2
+                            -
+                                $subtract:
+                                    -
+                                        $strLenBytes: '$quarter'
+                                    - 2
```
jmikola, line 47: Is the extra indentation the result of a script parsing the example code? If not, I think this can be more concise.

You can inline `{ $strLenBytes: '$quarter' }`, and `$subtract` shouldn't need a new line and indentation. We frequently do the following in spec tests:

```yaml
runOnRequirements:
  - minServerVersion: "4.0"
    topologies: [ replicaset ]
```

The same may apply to other files but I'll defer to you if it's worth addressing. I personally find the dashes for array elements on their own line a bit confusing when visually scanning a file.

generator/js2yaml.html
```diff
@@ -87,6 +89,10 @@ <h1>Convert JS examples into Yaml</h1>
         }
 
         if (Array.isArray(object)) {
+            if (object.length === 0) {
+                return ' []';
+			}
```
jmikola: Heads up: you're using tabs for indentation here and in the diff below. Looks like it should have been spaces.

tests/Builder/Expression/Pipelines.php
```diff
@@ -518,6 +518,27 @@ enum Pipelines: string
     ]
     JSON;
 
+    /**
+     * Example
```
jmikola, line 522: Do these names have any value? It looks like they're already incorporated into the `case` identifier (after a transformation of the base filename). Considering removing them in a subsequent PR. I agree the `@see` links are relevant, though.

Also, I forget if this ever came up in an earlier PR but it'd be good to check that the generator enforces unique test names per file -- unless you're intentionally relying on PHP to alert you to duplicate `case` IDs when failing to parse this generated file.

tests/Builder/Expression/RegexFindAllOperatorTest.php
```diff
@@ -0,0 +1,100 @@
...
+                    input: Expression::stringFieldPath('description'),
+                    regex: 'line',
+                    options: 'i',
+                ),
+            ),
+            Stage::addFields(
+                returnObject: Expression::regexFindAll(
+                    input: Expression::stringFieldPath('description'),
+                    regex: new Regex('line'),
+                    options: 'i',
```
jmikola, line 38: Should there be a manual test for the invalid input when the regex BSON type includes options and the `$regexFindAll` `options` option is also specified? Or do you rely on the server to catch that?

jmikola (commented):
Note the various Psalm errors.

generator/config/expression/substrBytes.yaml
GromNaN: I'm using a very standardized and exploded syntax for Yaml. It's generated by my `js2yaml.html` tool.

I can reindent all the Yaml with 2 spaces and remove the empty `-` lines in an other PR. PHPStorm can do this formatting for me.

generator/js2yaml.html
GromNaN: Thanks, fixed.
We may need a `.editorconfig` to have a cross-editor configuration if indentation style.

generator/config/expression/substrBytes.yaml
jmikola: Entirely up to you. This is the kind of formatting I try to avoid when writing spec tests, but I understand why a tool might produce it.

generator/config/expression/substrBytes.yaml
jmikola: This was a mistake on my part. Please ignore.

generator/js2yaml.html
```diff
@@ -16,16 +16,18 @@
...
   // Tests
   {
     $project: {
       date: new Date(36),
-	  regex: /^123/,
-	  regex_with_flags: /^123/i,
-	  binary: BinData(0, "ID=="),
+      regex: /^123/,
+      regex_with_flags: /^123/i,
+      binary: BinData(0, "ID=="),
```
jmikola:
```suggestion
      binary: BinData(0, "IA=="),
```

May as well use the canonical value (per [DOCSP-35733](https://jira.mongodb.org/browse/DOCSP-35733)).

tests/Builder/Expression/Pipelines.php
jmikola: @GromNaN ^^

jmikola (approved):
Several unaddressed comment threads but nothing significant. Mind the Psalm errors.

tests/Builder/Expression/Pipelines.php
GromNaN: It's all generated, there is no cost of having this names, and it may be more readable than the CamelCasedCaseName.

I rely on the php generator to check there is no duplicate name. Also PHP throw an error if there is a duplicate value in enum cases (benefit of using an enum over class constants).

tests/Builder/Expression/RegexFindAllOperatorTest.php
GromNaN: This would have the advantage of triggering an error early on instead of waiting for the command to be executed.
This check can be in the factory method, as it's the only place we can alter the code before generated code.

PR #43, opened 2024-01-22
Prefix Yaml tags with `!bson_`

GromNaN:
After @jmikola's comment: https://github.com/mongodb/mongo-php-builder/pull/37#discussion_r1461971431

### DIFF
--- a/generator/src/OperatorTestGenerator.php
+++ b/generator/src/OperatorTestGenerator.php
@@ -138,11 +138,11 @@ class OperatorTestGenerator extends OperatorGenerator
             $value = $object->getValue();
 
             return match ($object->getTag()) {
-                'regex' => new Regex(...(array) $value),
-                'long' => new Int64($value),
-                'double' => new Decimal128($value),
-                'date' => new UTCDateTime($value),
-                'binary' => new Binary(base64_decode($value)),
+                'bson_regex' => new Regex(...(array) $value),
+                'bson_long' => new Int64($value),
+                'bson_double' => new Decimal128($value),
+                'bson_date' => new UTCDateTime($value),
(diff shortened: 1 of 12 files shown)


generator/src/OperatorTestGenerator.php
```diff
@@ -138,11 +138,11 @@ private function convertYamlTaggedValues(mixed $object): mixed
...
 
             return match ($object->getTag()) {
-                'regex' => new Regex(...(array) $value),
-                'long' => new Int64($value),
-                'double' => new Decimal128($value),
-                'date' => new UTCDateTime($value),
-                'binary' => new Binary(base64_decode($value)),
+                'bson_regex' => new Regex(...(array) $value),
+                'bson_long' => new Int64($value),
+                'bson_double' => new Decimal128($value),
```
jmikola: If we're prefixing these with `bson_`, can we use the actual names for clarity?

That would be `bson_int64` above, `bson_decimal128` here, and `bson_utcdatetime` below.

jmikola (changes requested):

generator/src/OperatorTestGenerator.php
GromNaN: Are this names shared with other drivers ? The idea is also that this config files could be extracted and used by other languages to create agg builders.

jmikola: These names are consistent with the [BSON spec](https://bsonspec.org/spec.html), which pre-dates any driver specs. That's the reason we chose the names when implementing `ext-mongodb` (vs. keeping what we originally had in `ext-mongo`).

GromNaN: Ok, renamed.

jmikola (approved):

PR #46, opened 2024-01-23
PHPLIB-1250 Split encoders and fix psalm issues

alcaeus:
PHPLIB-1250

This PR splits the main `BuilderEncoder` into multiple files. This allows for more isolation of separate code (which I'm sure can be improved further), but also allows for easy introduction of custom codecs (currently not tested yet).

In addition, this PR cleans up a number of psalm issues and adds other issues to the baseline. I'll comment on those as necessary.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -29,7 +29,15 @@
       <code>$val</code>
     </MixedAssignment>
   </file>
+  <file src="src/Builder/Encoder/OutputWindowEncoder.php">
+    <MixedArgument>
+      <code>$result</code>
+    </MixedArgument>
+  </file>
   <file src="src/Builder/Encoder/QueryEncoder.php">
+    <MixedArgument>
+      <code><![CDATA[$this->recursiveEncode($value)]]></code>
+    </MixedArgument>
     <MixedAssignment>
       <code>$subValue</code>
       <code>$value</code>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,18 +1,22 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <files psalm-version="5.20.0@3f284e96c9d9be6fe6b15c79416e1d1903dcfef4">
-  <file src="src/Builder/BuilderEncoder.php">
-    <MixedArgument>
-      <code>$value</code>
-    </MixedArgument>
+  <file src="src/Builder/Encoder/AbstractExpressionEncoder.php">
+    <MixedAssignment>
+      <code>$val</code>
+      <code>$val</code>
+      <code>$value[$key]</code>
+    </MixedAssignment>
   </file>
   <file src="src/Builder/Encoder/CombinedFieldQueryEncoder.php">
     <MixedAssignment>
       <code>$filter</code>
       <code>$filterValue</code>
-      <code>$val</code>
-      <code>$val</code>
-      <code>$value[$key]</code>
     </MixedAssignment>
+  </file>
+  <file src="src/Builder/Encoder/FieldPathEncoder.php">
+    <NoInterfaceProperties>
+      <code><![CDATA[$value->name]]></code>
+    </NoInterfaceProperties>
   </file>
   <file src="src/Builder/Encoder/OperatorEncoder.php">
     <MixedAssignment>
@@ -23,46 +27,103 @@
       <code>$val</code>
       <code>$val</code>
       <code>$val</code>
-      <code>$val</code>
-      <code>$val</code>
-      <code>$value[$key]</code>
     </MixedAssignment>
-  </file>
-  <file src="src/Builder/Encoder/OutputWindowEncoder.php">
-    <MixedAssignment>
-      <code>$val</code>
-      <code>$val</code>
-      <code>$value[$key]</code>
-    </MixedAssignment>
-  </file>
-  <file src="src/Builder/Encoder/PipelineEncoder.php">
-    <MixedAssignment>
-      <code>$encoded[]</code>
-    </MixedAssignment>
+    <UndefinedConstant>
+      <code>$value::ENCODE</code>
+    </UndefinedConstant>
   </file>
   <file src="src/Builder/Encoder/QueryEncoder.php">
-    <MixedArgument>
-      <code>$key</code>
-      <code><![CDATA[$this->recursiveEncode($value)]]></code>
-    </MixedArgument>
     <MixedAssignment>
-      <code>$key</code>
       <code>$subValue</code>
-      <code>$val</code>
-      <code>$val</code>
       <code>$value</code>
-      <code>$value[$key]</code>
     </MixedAssignment>
   </file>
-  <file src="src/Builder/Type/CombinedFieldQuery.php">
-    <MixedAssignment>
-      <code>$fieldQuery</code>
-    </MixedAssignment>
+  <file src="src/Builder/Projection/ElemMatchOperator.php">
+    <MixedArgumentTypeCoercion>
+      <code>$query</code>
+    </MixedArgumentTypeCoercion>
+  </file>
+  <file src="src/Builder/Query.php">
+    <ArgumentTypeCoercion>
+      <code>$query</code>
+    </ArgumentTypeCoercion>
+  </file>
+  <file src="src/Builder/Query/ElemMatchOperator.php">
+    <MixedArgumentTypeCoercion>
+      <code>$query</code>
+    </MixedArgumentTypeCoercion>
+  </file>
+  <file src="src/Builder/Stage/AddFieldsStage.php">
+    <PropertyTypeCoercion>
+      <code>$expression</code>
+    </PropertyTypeCoercion>
+    <TooManyTemplateParams>
+      <code>stdClass</code>
+    </TooManyTemplateParams>
+  </file>
+  <file src="src/Builder/Stage/FacetStage.php">
+    <PropertyTypeCoercion>
+      <code>$facet</code>
+    </PropertyTypeCoercion>
+    <TooManyTemplateParams>
+      <code>stdClass</code>
+    </TooManyTemplateParams>
+  </file>
+  <file src="src/Builder/Stage/GeoNearStage.php">
+    <MixedArgumentTypeCoercion>
+      <code>$query</code>
+    </MixedArgumentTypeCoercion>
+  </file>
+  <file src="src/Builder/Stage/GraphLookupStage.php">
+    <MixedArgumentTypeCoercion>
+      <code>$restrictSearchWithMatch</code>
+    </MixedArgumentTypeCoercion>
+  </file>
+  <file src="src/Builder/Stage/GroupStage.php">
+    <PropertyTypeCoercion>
+      <code>$field</code>
+    </PropertyTypeCoercion>
+    <TooManyTemplateParams>
+      <code>stdClass</code>
+    </TooManyTemplateParams>
+  </file>
+  <file src="src/Builder/Stage/MatchStage.php">
+    <MixedArgumentTypeCoercion>
+      <code>$query</code>
+    </MixedArgumentTypeCoercion>
+  </file>
+  <file src="src/Builder/Stage/ProjectStage.php">
+    <PropertyTypeCoercion>
+      <code>$specification</code>
+    </PropertyTypeCoercion>
+    <TooManyTemplateParams>
+      <code>stdClass</code>
+    </TooManyTemplateParams>
+  </file>
+  <file src="src/Builder/Stage/SetStage.php">
+    <PropertyTypeCoercion>
+      <code>$field</code>
+    </PropertyTypeCoercion>
+    <TooManyTemplateParams>
+      <code>stdClass</code>
+    </TooManyTemplateParams>
+  </file>
+  <file src="src/Builder/Type/OutputWindow.php">
+    <DocblockTypeContradiction>
+      <code><![CDATA[! is_string($documents[1]) && ! is_int($documents[1])]]></code>
+      <code><![CDATA[! is_string($range[1]) && ! is_numeric($range[1])]]></code>
+    </DocblockTypeContradiction>
   </file>
   <file src="src/Builder/Type/QueryObject.php">
     <MixedAssignment>
       <code>$queries[$fieldPath]</code>
       <code>$query</code>
     </MixedAssignment>
+    <RedundantConditionGivenDocblockType>
+      <code><![CDATA[count($queriesOrArrayOfQueries) === 1 &&
+            isset($queriesOrArrayOfQueries[0]) &&
+            is_array($queriesOrArrayOfQueries[0]) &&
+            count($queriesOrArrayOfQueries[0]) > 0]]></code>
+    </RedundantConditionGivenDocblockType>
   </file>
 </files>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -28,9 +28,6 @@
       <code>$val</code>
       <code>$val</code>
     </MixedAssignment>
-    <UndefinedConstant>
-      <code>$value::ENCODE</code>
-    </UndefinedConstant>
   </file>
   <file src="src/Builder/Encoder/QueryEncoder.php">
     <MixedAssignment>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,3 +1,68 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="5.15.0@5c774aca4746caf3d239d9c8cadb9f882ca29352">
+<files psalm-version="5.20.0@3f284e96c9d9be6fe6b15c79416e1d1903dcfef4">
+  <file src="src/Builder/BuilderEncoder.php">
+    <MixedArgument>
+      <code>$value</code>
+    </MixedArgument>
+  </file>
+  <file src="src/Builder/Encoder/CombinedFieldQueryEncoder.php">
+    <MixedAssignment>
+      <code>$filter</code>
+      <code>$filterValue</code>
+      <code>$val</code>
+      <code>$val</code>
+      <code>$value[$key]</code>
+    </MixedAssignment>
+  </file>
+  <file src="src/Builder/Encoder/OperatorEncoder.php">
+    <MixedAssignment>
+      <code>$result</code>
+      <code>$result[]</code>
+      <code>$val</code>
+      <code>$val</code>
+      <code>$val</code>
+      <code>$val</code>
+      <code>$val</code>
+      <code>$val</code>
+      <code>$val</code>
+      <code>$value[$key]</code>
+    </MixedAssignment>
+  </file>
+  <file src="src/Builder/Encoder/OutputWindowEncoder.php">
+    <MixedAssignment>
+      <code>$val</code>
+      <code>$val</code>
+      <code>$value[$key]</code>
+    </MixedAssignment>
+  </file>
+  <file src="src/Builder/Encoder/PipelineEncoder.php">
+    <MixedAssignment>
+      <code>$encoded[]</code>
+    </MixedAssignment>
+  </file>
+  <file src="src/Builder/Encoder/QueryEncoder.php">
+    <MixedArgument>
+      <code>$key</code>
+      <code><![CDATA[$this->recursiveEncode($value)]]></code>
+    </MixedArgument>
+    <MixedAssignment>
+      <code>$key</code>
+      <code>$subValue</code>
+      <code>$val</code>
+      <code>$val</code>
+      <code>$value</code>
+      <code>$value[$key]</code>
+    </MixedAssignment>
+  </file>
+  <file src="src/Builder/Type/CombinedFieldQuery.php">
+    <MixedAssignment>
+      <code>$fieldQuery</code>
+    </MixedAssignment>
+  </file>
+  <file src="src/Builder/Type/QueryObject.php">
+    <MixedAssignment>
+      <code>$queries[$fieldPath]</code>
+      <code>$query</code>
+    </MixedAssignment>
+  </file>
 </files>
--- a/src/Builder/BuilderEncoder.php
+++ b/src/Builder/BuilderEncoder.php
@@ -60,16 +60,14 @@ class BuilderEncoder implements Encoder
             return false;
         }
 
-        $encoder = $this->getEncoderFor($value);
-
-        return $encoder !== null && $encoder->canEncode($value);
+        return (bool) $this->getEncoderFor($value)?->canEncode($value);
     }
 
     public function encode(mixed $value): stdClass|array|string
     {
         $encoder = $this->getEncoderFor($value);
 
-        if (! $encoder || ! $encoder->canEncode($value)) {
+        if (! $encoder?->canEncode($value)) {
             throw UnsupportedValueException::invalidEncodableValue($value);
         }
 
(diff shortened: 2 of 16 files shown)


psalm-baseline.xml
```diff
@@ -1,3 +1,129 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="5.15.0@5c774aca4746caf3d239d9c8cadb9f882ca29352">
+<files psalm-version="5.20.0@3f284e96c9d9be6fe6b15c79416e1d1903dcfef4">
+  <file src="src/Builder/Encoder/AbstractExpressionEncoder.php">
+    <MixedAssignment>
```
alcaeus, line 4: A lot of `MixedAssignment` calls are the result of incomplete type annotations throughout the generated files.

psalm-baseline.xml
```diff
@@ -1,3 +1,129 @@
...
+    </MixedAssignment>
+  </file>
+  <file src="src/Builder/Encoder/CombinedFieldQueryEncoder.php">
+    <MixedAssignment>
+      <code>$filter</code>
+      <code>$filterValue</code>
+    </MixedAssignment>
+  </file>
+  <file src="src/Builder/Encoder/FieldPathEncoder.php">
+    <NoInterfaceProperties>
```
alcaeus, line 17: Fixing this involves creating a new generator for field path classes so we can add a `getName` interface methods. It irks me that public properties cannot be part of the interface contract, but here we are.

psalm-baseline.xml
```diff
@@ -1,3 +1,129 @@
...
+    <MixedAssignment>
+      <code>$result</code>
+      <code>$result[]</code>
+      <code>$val</code>
+      <code>$val</code>
+      <code>$val</code>
+      <code>$val</code>
+      <code>$val</code>
+    </MixedAssignment>
+    <UndefinedConstant>
```
alcaeus: Similar to the interface property above, this undefined constant stems from us checking for an interface, then accessing a class constant.

psalm-baseline.xml
```diff
@@ -1,3 +1,129 @@
...
+    </UndefinedConstant>
+  </file>
+  <file src="src/Builder/Encoder/QueryEncoder.php">
+    <MixedAssignment>
+      <code>$subValue</code>
+      <code>$value</code>
+    </MixedAssignment>
+  </file>
+  <file src="src/Builder/Projection/ElemMatchOperator.php">
+    <MixedArgumentTypeCoercion>
```
alcaeus, line 47: `MixedArgumentTypeCoercion` errors are due to the query expression not using the correct array element types but instead using the `array` type. I tried to fix this, but the code generator doesn't accept a type like `array<int|string>`. This will have to be looked at separately.

psalm-baseline.xml
```diff
@@ -1,3 +1,129 @@
...
+  <file src="src/Builder/Query/ElemMatchOperator.php">
+    <MixedArgumentTypeCoercion>
+      <code>$query</code>
+    </MixedArgumentTypeCoercion>
+  </file>
+  <file src="src/Builder/Stage/AddFieldsStage.php">
+    <PropertyTypeCoercion>
+      <code>$expression</code>
+    </PropertyTypeCoercion>
+    <TooManyTemplateParams>
```
alcaeus, line 65: We use `stdClass<foo|bar>` to indicate an object with an unknown number of properties of a specific type. However, `stdClass` and `object` cannot be templated this way, which is why we're seeing this error. The `stdClass{foo: int}` syntax doesn't work for us here as we don't know the keys of the object, but only the expected value types. I've asked in the Symfony Slack to get more info about this and will update once I know more.

src/Builder/Encoder/PipelineEncoder.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    /** @template-use EncodeIfSupported<array, Pipeline> */
+    use EncodeIfSupported;
+
+    public function canEncode(mixed $value): bool
+    {
+        return $value instanceof Pipeline;
+    }
+
+    public function encode(mixed $value): stdClass|array|string
```
GromNaN: This returns always a list.

```suggestion
    public function encode(mixed $value): array
```

psalm-baseline.xml
```diff
@@ -1,3 +1,129 @@
...
+  <file src="src/Builder/Stage/SetStage.php">
+    <PropertyTypeCoercion>
+      <code>$field</code>
+    </PropertyTypeCoercion>
+    <TooManyTemplateParams>
+      <code>stdClass</code>
+    </TooManyTemplateParams>
+  </file>
+  <file src="src/Builder/Type/OutputWindow.php">
+    <DocblockTypeContradiction>
```
alcaeus, line 117: This and the `RedundantConditionGivenDocblockType` checks are psalm telling us that the condition is impossible due to the docblock. However, since we're being extra cautious here and any user not using psalm is able to send any types they want, I think it's fair to keep this around unless we want to accept potential fatal errors when users abuse this.

src/Builder/Encoder/PipelineEncoder.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        if (! $this->canEncode($value)) {
```
GromNaN, line 26: We need a method `encodeIfSupported` to avoid repeating this 3 lines in all encoders.

src/Builder/BuilderEncoder.php
```diff
@@ -4,323 +4,101 @@
...
-            }
-
-            foreach ($filter as $key => $value) {
-                $result->{$key} = $value;
-            }
-        }
+        $encoder = $this->getEncoderFor($value);
 
-        return $result;
+        return $encoder !== null && $encoder->canEncode($value);
```
GromNaN: There is syntax for that:

```php
return (bool) $this->getEncoderFor($value)?->canEncode($value);
```

src/Builder/BuilderEncoder.php
```diff
@@ -4,323 +4,101 @@
...
-                }
-            } else {
-                if (property_exists($result, (string) $key)) {
-                    throw new LogicException(sprintf('Duplicate key "%s" in query object', $key));
-                }
-
-                $result->{$key} = $this->encodeIfSupported($value);
-            }
+        if (! $encoder || ! $encoder->canEncode($value)) {
+            throw UnsupportedValueException::invalidEncodableValue($value);
```
GromNaN, line 72: The `$encoder->canEncode($value)` part is already checked inside `$encoder->encode($value)`.

src/Builder/Type/QueryObject.php
```diff
@@ -27,7 +25,7 @@ final class QueryObject implements QueryInterface
 {
     public readonly array $queries;
 
-    /** @param array<QueryInterface|FieldQueryInterface|Decimal128|Int64|Regex|stdClass|array|bool|float|int|string|null> $queries */
+    /** @param array<QueryInterface|FieldQueryInterface|Type|stdClass|array|bool|float|int|string|null> $queries */
```
jmikola, line 28: Is this a preemptive change so we don't need to keep adding specific BSON classes as the library is developed, or is it already possible that this method would receive additional BSON types?

src/Builder/Encoder/PipelineEncoder.php
jmikola: In this case, would `list` be preferable?

src/Builder/Encoder/PipelineEncoder.php
jmikola: Isn't `encodeIfSupported()` already provided via the EncodeIfSupported trait above? The trait implementation utilizes the `canEncode()` and `encode()` methods.

I understand that doesn't address the repetition of checking `canEncode()` in each `encode()` method, though.

psalm-baseline.xml
jmikola: According to [Object Interfaces: Constants](https://www.php.net/manual/en/language.oop5.interfaces.php#language.oop5.interfaces.constants), PHP 8.1 allows overriding constants. Is there any reason Psalm cannot support this?

Or does it stem from OperatorInterface not actually defining the constant, and it just being a convention in this library for implementing classes to do so? If so, perhaps we need to incorporate a new method for the encoding directive?

src/Builder/BuilderEncoder.php
```diff
@@ -4,323 +4,101 @@
...
-                $filter = get_object_vars($filter);
-            } elseif (! is_array($filter)) {
-                throw new LogicException(sprintf('Query filters must an array or an object. Got "%s"', get_debug_type($filter)));
-            }
-
-            foreach ($filter as $key => $value) {
-                $result->{$key} = $value;
-            }
-        }
+        $encoder = $this->getEncoderFor($value);
```
jmikola: This reminds me of the original implementation of codec libraries (https://github.com/mongodb/mongo-php-library/pull/1160), which were removed from PHPLIB before release.

src/Builder/BuilderEncoder.php
```diff
@@ -4,323 +4,101 @@
...
-     * Nested arrays and objects must be encoded recursively.
-     */
-    private function recursiveEncode(mixed $value): mixed
-    {
-        if (is_array($value)) {
-            foreach ($value as $key => $val) {
-                $value[$key] = $this->recursiveEncode($val);
+        // First attempt: match class name exactly
+        foreach ($encoderList as $className => $encoderClass) {
+            if ($className == $valueClass) {
```
jmikola: Any reason not to use `===` here?

src/Builder/BuilderEncoder.php
```diff
@@ -4,323 +4,101 @@
...
-            return $value;
         }
 
-        if ($value instanceof stdClass) {
-            foreach (get_object_vars($value) as $key => $val) {
-                $value->{$key} = $this->recursiveEncode($val);
+        // Second attempt: catch child classes
+        foreach ($encoderList as $className => $encoderClass) {
+            if ($value instanceof $className) {
+                return $this->cachedEncoders[$valueClass] = new $encoderClass($this);
```
jmikola, line 95: This will construct encoder instances for each subclass. Do you want to leave a comment about potentially restructuring this to be more efficient?

Likewise, you can also optimize for the miss case. It's not as important for `encode()` since that's an error condition but would help with `canEncode()` calls if those are more frequent.

Maybe this is something to ticket for potential profiling and refactoring at a later point.

src/Builder/Encoder/AbstractExpressionEncoder.php
```diff
@@ -0,0 +1,53 @@
...
+    final public function __construct(protected readonly BuilderEncoder $encoder)
+    {
+    }
+
+    /**
+     * Nested arrays and objects must be encoded recursively.
+     *
+     * @psalm-param T $value
+     *
+     * @psalm-return (T is object ? object : (T is array ? array : mixed))
```
jmikola: Should this use `stdClass` instead of `object`? [Object Types](https://psalm.dev/docs/annotating_code/type_syntax/object_types/) suggests that those are distinct.

That said, the code will still end up returning a non-stdClass object as-is, which won't invalidate the this annotation.

Separate question: why return `mixed` instead of `T` for the non-object, non-array case? [`@template`, `@psalm-template`](https://psalm.dev/docs/annotating_code/templated_annotations/#template-psalm-template) has an example of doing so, and I assume all extending classes will substitute `T` (correct me if I'm mistaken).

src/Builder/Encoder/CombinedFieldQueryEncoder.php
```diff
@@ -0,0 +1,52 @@
...
+{
+    /** @template-use EncodeIfSupported<stdClass, CombinedFieldQuery> */
+    use EncodeIfSupported;
+
+    public function canEncode(mixed $value): bool
+    {
+        return $value instanceof CombinedFieldQuery;
+    }
+
+    public function encode(mixed $value): stdClass
```
jmikola, line 30: Noted that this was lifted from `BuilderEncoder::encodeCombinedFilter()`, so I'm not going to spend time reviewing this (and similar) methods.

src/Builder/Encoder/FieldPathEncoder.php
```diff
@@ -0,0 +1,31 @@
...
+        return $value instanceof FieldPathInterface;
+    }
+
+    public function encode(mixed $value): string
+    {
+        if (! $this->canEncode($value)) {
+            throw UnsupportedValueException::invalidEncodableValue($value);
+        }
+
+        // TODO: needs method because of interface
```
jmikola: Noted that this was extracted from the inline logic in BuilderEncode. Should this TODO be ticketed?

src/Builder/Encoder/OperatorEncoder.php
```diff
@@ -0,0 +1,161 @@
...
+        return $value instanceof OperatorInterface;
+    }
+
+    public function encode(mixed $value): stdClass
+    {
+        if (! $this->canEncode($value)) {
+            throw UnsupportedValueException::invalidEncodableValue($value);
+        }
+
+        switch ($value::ENCODE) {
```
jmikola, line 41: Related to an earlier question in `psalm-baseline.xml`, but should OperatorInterface have a method that returns the encoding type instead of utilizing a class constant?

Alternatively, if constants are being used for performance reasons, then can we define a constant with some invalid sentinel value on the base interface and require implementing classes to override it?

src/Builder/Encoder/OperatorEncoder.php
```diff
@@ -0,0 +1,161 @@
...
+    /**
+     * Encode the value as an array of properties, in the order they are defined in the class.
+     */
+    private function encodeAsArray(OperatorInterface $value): stdClass
+    {
+        $result = [];
+        /** @var mixed $val */
+        foreach (get_object_vars($value) as $val) {
+            // Skip optional arguments.
+            // $slice operator has the optional <position> argument in the middle of the array
```
jmikola: @GromNaN: This is unrelated to the PR, but is the `$slice` expression operator still the only use case for `Optional::Undefined` in an `encodeAsArray()` context?

jmikola:
```suggestion
            // Skip optional arguments. For example, the $slice expression operator has an optional <position> argument
            // in the middle of the array.
```

Rewriting this for clarity. I forgot that it was the _expression_ operator. Alternatively, we could refer to the actual `MongoDB\Builder\Expression\SliceOperator` class since that's where `Optional::Undefined` is used, but it'd be less concise.

src/Builder/Encoder/OperatorEncoder.php
```diff
@@ -0,0 +1,161 @@
...
+            if ($val === Optional::Undefined) {
+                continue;
+            }
+
+            $val = $this->recursiveEncode($val);
+
+            if ($key === 'geometry') {
+                if (is_object($val) && property_exists($val, '$geometry')) {
+                    $result->{'$geometry'} = $val->{'$geometry'};
+                } elseif (is_array($val) && array_key_exists('$geometry', $val)) {
```
jmikola, line 127: Seeing these side by side, I just realized that `property_exists()` and `array_key_exists()` use inconsistent argument ordering. Oh, PHP.

src/Builder/Encoder/OperatorEncoder.php
```diff
@@ -0,0 +1,161 @@
...
+
+        throw new LogicException(sprintf('Class "%s" does not have a single property.', $value::class));
+    }
+
+    private function wrap(OperatorInterface $value, mixed $result): stdClass
+    {
+        $object = new stdClass();
+        $object->{$value->getOperator()} = $result;
+
+        return $object;
```
jmikola, line 159: Do we want a one-liner for this?

```suggestion
        return (object) [$value->getOperator() => $result];
```

src/Builder/Encoder/QueryEncoder.php
```diff
@@ -0,0 +1,57 @@
...
+                // The sub-objects is merged into the main object, replacing duplicate keys
+                foreach (get_object_vars($this->recursiveEncode($value)) as $subKey => $subValue) {
+                    if (property_exists($result, $subKey)) {
+                        throw new LogicException(sprintf('Duplicate key "%s" in query object', $subKey));
+                    }
+
+                    $result->{$subKey} = $subValue;
+                }
+            } else {
+                if (property_exists($result, (string) $key)) {
```
jmikola, line 47: Is the string-cast necessary because `QueryObject::$queries` may be a list? I'm looking at its constructor now and see several ways of populating the class property.

src/Builder/Encoder/VariableEncoder.php
```diff
@@ -0,0 +1,30 @@
...
+        return $value instanceof Variable;
+    }
+
+    public function encode(mixed $value): string
+    {
+        if (! $this->canEncode($value)) {
+            throw UnsupportedValueException::invalidEncodableValue($value);
+        }
+
+        return '$$' . $value->name;
```
jmikola, line 29: In FieldPathEncoder you have the following:

```
// TODO: needs method because of interface
```

Is this also affected? If so, can we cover both in a single JIRA ticket?

src/Builder/Type/QueryObject.php
GromNaN: This type is inherited from code generator. We accept any BSON Serializable, as they implement BSON Type.

src/Builder/Encoder/QueryEncoder.php
GromNaN: Field names can be numeric strings, which are automatically casted to int by PHP when used as array-key. And the function expects a string.

src/Builder/Encoder/OperatorEncoder.php
GromNaN: That's the only case I've found, for now.

src/Builder/Encoder/OperatorEncoder.php
GromNaN: I'd like to microbench both implementations.

psalm-baseline.xml
alcaeus: Removed this entry and added a constant to `OperatorInterface`.

src/Builder/Encoder/PipelineEncoder.php
alcaeus: I changed the return type to `array`, and the resulting type to `list<array|stdClass|string>` as `list` itself wasn't sufficient. Since we don't have `BuilderEncoder::encode` typed any more strictly, this is the best we can do.

src/Builder/BuilderEncoder.php
alcaeus: Updated, along with a later check.

src/Builder/BuilderEncoder.php
alcaeus: Yes, however there is no guarantee that somebody will never call `encode` with an invalid value, as the contract does not require calling `canEncode` for a value beforehand.

We could solve this with the introduction of a `null` type for the `$value` argument in the `Encoder::encode` definition, but that would require us to always accept `null` values in any encoder, which might still necessitate calling `canEncode`. `void` as PHP's true bottom type is unfortunately not available as an argument type, so there's no proper typing way around this.

src/Builder/BuilderEncoder.php
alcaeus: Yes. A primary concern in that implementation was iterating the list of encoders and calling `canEncode` on each of them. Here we have a list of known encoders and the classes they should be able to encode, so we can decide on a codec upfront.

src/Builder/BuilderEncoder.php
alcaeus: A missing phpcs rule is the reason. I've enabled the rule and fixed this occurrence.

src/Builder/Encoder/AbstractExpressionEncoder.php
alcaeus: Updated to use `stdClass` instead of `object`. If the type is neither `array` nor `stdClass`, we can't make any further assumptions about the return type as we don't know what the `encodeIfSupported` method will return for a type `T`. I'll have to play around with psalm to see if this can actually be done in a dynamic list of encoders.

src/Builder/Encoder/FieldPathEncoder.php
alcaeus: I'll defer to @GromNaN on this since this may involve more work but also isn't bad to cause a fatal error in case we forget it somewhere.

src/Builder/Encoder/OperatorEncoder.php
alcaeus: A method would be an alternative, but for the time being I've added the constant to `OperatorInterface`. With a default value of `Encode::Single`, the effect of not overriding this constant would be failing tests for any operator that doesn't encode as single operator (which the majority of our operators do).

src/Builder/Encoder/OperatorEncoder.php
alcaeus: Deferring to @GromNaN for this

src/Builder/Encoder/PipelineEncoder.php
alcaeus: Correction: it's `list<mixed>`, as `encodeIfSupported` is used throughout the encoders, which results in an effective return type of `mixed` whenever we defer to the `BuilderEncoder` class.

src/Builder/Encoder/OperatorEncoder.php
alcaeus: Since I was already microbenching `array_key_exists`, I also checked this. There's no significant performance difference (30 ms when testing 1M object creations), so there's no performance benefit in using either.

GromNaN (approved):

src/Builder/Encoder/VariableEncoder.php
jmikola: @alcaeus: :point_up:

src/Builder/Type/OperatorInterface.php
```diff
@@ -9,5 +9,8 @@
  */
 interface OperatorInterface
 {
+    /** To be overridden by implementing classes */
+    public const ENCODE = Encode::Single;
```
jmikola: Noted that you're defaulting to `Single` here instead of introducing an invalid case. My thinking was the invalid case would make it easier to catch mistakes when failing to override this, since `Single` would just happen to work (at least for encoding) provided the class only had a single property.

jmikola (approved):
LGTM, but not my outstanding question about TODO comments.

src/Builder/Type/OperatorInterface.php
GromNaN: You can add a `Undefined` case to the enum. We could also use `null`, but the const may be typed one day.

jmikola: I'm personally a fan of `Undefined` but won't push for it.

src/Builder/BuilderEncoder.php
alcaeus: Forgot to mention that I refactored the logic:
a) `null` values for classes with no encoders are now cached
b) The "first" attempt section now uses `isset` instead of looping through the list to check the key (🤦‍♂️)

src/Builder/Encoder/FieldPathEncoder.php
alcaeus: Created PHPLIB-1379 to track this.

src/Builder/Encoder/VariableEncoder.php
alcaeus: Tracked in [PHPLIB-1379](https://jira.mongodb.org/browse/PHPLIB-1379).

src/Builder/Type/OperatorInterface.php
alcaeus: Introduced `Undefined` value to the enum. Agree with not abusing `null` for this to keep the constant strictly typed.

PR #48, opened 2024-01-23
PHPLIB-1357 Add tests on Trigonometry Expression Operators

GromNaN:
Fix [PHPLIB-1357](https://jira.mongodb.org/browse/PHPLIB-1357)

https://www.mongodb.com/docs/manual/reference/operator/aggregation/#trigonometry-expression-operators

- `$sin`
- `$cos`
- `$tan`
- `$asin`
- `$acos`
- `$atan`
- `$atan2`
- `$asinh`
- `$acosh`
- `$atanh`
- `$sinh`
- `$cosh`
- `$tanh`
- `$degreesToRadians`
- `$radiansToDegrees`

### DIFF
--- a/tests/Builder/Expression/AcoshOperatorTest.php
+++ b/tests/Builder/Expression/AcoshOperatorTest.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Tests\Builder\Expression;
+
+use MongoDB\Builder\Expression;
+use MongoDB\Builder\Pipeline;
+use MongoDB\Builder\Stage;
+use MongoDB\Tests\Builder\PipelineTestCase;
+
+/**
+ * Test $acosh expression
+ */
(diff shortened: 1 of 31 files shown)


tests/Builder/Expression/AcoshOperatorTest.php
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Test $acosh expression
+ */
+class AcoshOperatorTest extends PipelineTestCase
+{
+    public function testExample(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::addFields(
+                ...[
```
jmikola, line 21: Is this only necessary because `y-coordinate` is an invalid parameter name?

jmikola (approved):

tests/Builder/Expression/AcoshOperatorTest.php
GromNaN: Yes. The name can't be used as named argument.

PR #49, opened 2024-01-23
PHPLIB-1358 Add tests on Type Expression Operators

GromNaN:
Fix [PHPLIB-1358](https://jira.mongodb.org/browse/PHPLIB-1358)

https://www.mongodb.com/docs/manual/reference/operator/aggregation/#type-expression-operators

- `$convert`
- `$isNumber` fixed, as it get a single value
- `$toBool`
- ~`$toDate`~ already done
- `$toDecimal`
- `$toDouble`
- `$toInt`
- `$toLong`
- `$toObjectId`
- ~`$toString`~ already done
- `$type`

### DIFF
--- a/src/Builder/Expression/FactoryTrait.php
+++ b/src/Builder/Expression/FactoryTrait.php
@@ -970,14 +970,13 @@ trait FactoryTrait
      * New in MongoDB 4.4.
      *
      * @see https://www.mongodb.com/docs/manual/reference/operator/aggregation/isNumber/
-     * @no-named-arguments
-     * @param ExpressionInterface|Type|array|bool|float|int|non-empty-string|null|stdClass ...$expression
+     * @param ExpressionInterface|Type|array|bool|float|int|non-empty-string|null|stdClass $expression
      */
     public static function isNumber(
-        Type|ExpressionInterface|stdClass|array|bool|float|int|null|string ...$expression,
+        Type|ExpressionInterface|stdClass|array|bool|float|int|null|string $expression,
     ): IsNumberOperator
     {
-        return new IsNumberOperator(...$expression);
+        return new IsNumberOperator($expression);
     }
 
     /**
--- a/tests/Builder/Expression/ToDoubleOperatorTest.php
+++ b/tests/Builder/Expression/ToDoubleOperatorTest.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Tests\Builder\Expression;
+
+use MongoDB\Builder\Expression;
+use MongoDB\Builder\Pipeline;
+use MongoDB\Builder\Stage;
+use MongoDB\Tests\Builder\PipelineTestCase;
+
+/**
+ * Test $toDouble expression
+ */
+class ToDoubleOperatorTest extends PipelineTestCase
+{
+    public function testExample(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::addFields(
+                degrees: Expression::toDouble(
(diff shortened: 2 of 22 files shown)


src/Builder/Expression/FactoryTrait.php
```diff
@@ -970,14 +970,13 @@ public static function isArray(
      * New in MongoDB 4.4.
      *
      * @see https://www.mongodb.com/docs/manual/reference/operator/aggregation/isNumber/
-     * @no-named-arguments
-     * @param ExpressionInterface|Type|array|bool|float|int|non-empty-string|null|stdClass ...$expression
+     * @param ExpressionInterface|Type|array|bool|float|int|non-empty-string|null|stdClass $expression
      */
     public static function isNumber(
-        Type|ExpressionInterface|stdClass|array|bool|float|int|null|string ...$expression,
+        Type|ExpressionInterface|stdClass|array|bool|float|int|null|string $expression,
```
jmikola, line 976: What triggered both changes to the generated `FactoryTrait.php` and `IsNumberOperator.php` files?

The only modification to `isNumber.yaml` was new tests.

tests/Builder/Expression/ToDoubleOperatorTest.php
```diff
@@ -0,0 +1,33 @@
...
+
+use MongoDB\Builder\Expression;
+use MongoDB\Builder\Pipeline;
+use MongoDB\Builder\Stage;
+use MongoDB\Tests\Builder\PipelineTestCase;
+
+/**
+ * Test $toDouble expression
+ */
+class ToDoubleOperatorTest extends PipelineTestCase
```
jmikola, line 15: This is outside the scope of the PR, but why does the parameter type for `ToDoubleOperator::__construct()` include `non-empty-string` instead of just `string`? In `toDouble.yaml`, the `$expression` argument has a type `expression`.

I traced this to a "strings cannot be empty" comment in `OperatorGenerator::getAcceptedTypes()`, but it's unclear to me why that is. That's also the case for StrLenBytesOperator, which I'd expect would be able to take a string of any length.

src/Builder/Expression/FactoryTrait.php
GromNaN: I removed `variadic: array` for this operator. It accepts only a single value.
That's an actual bug fix.

tests/Builder/Expression/ToDoubleOperatorTest.php
GromNaN: Out of zeal, I've put this type for all the strings, but that's probably overkill.

src/Builder/Expression/FactoryTrait.php
jmikola: Ah, I missed that when reviewing the diff. Thanks for clarifying!

tests/Builder/Expression/ToDoubleOperatorTest.php
jmikola: I feel like there would be at least _some_ cases where an empty string is a legitimate argument for some operators. If you agree, consider opening a separate issue to investigate that at a later point.

jmikola (approved):
LGTM with suggestion to revisit the `non-empty-string` typing at a later point.

tests/Builder/Expression/ToDoubleOperatorTest.php
GromNaN: Tracked in [PHPLIB-1373](https://jira.mongodb.org/browse/PHPLIB-1373)

PR #51, opened 2024-01-24
PHPLIB-1362 Add tests on Window Operators

GromNaN:
Fix [PHPLIB-1362](https://jira.mongodb.org/browse/PHPLIB-1362)

https://www.mongodb.com/docs/manual/reference/operator/aggregation/#window-operators

- ~`$addToSet`~
- ~`$avg`~
- ~`$bottom`~
- ~`$bottomN`~
- ~`$count`~
- `$covariancePop`
- `$covarianceSamp`
- `$denseRank`
- `$derivative`
- `$documentNumber`
- `$expMovingAvg`
- ~`$first`~
- `$integral`
- ~`$last`~
- `$linearFill`
- `$locf`
- ~`$max`~
- ~`$min`~
- ~`$minN`~
- ~`$push`~
- `$rank`
- `$shift`
- ~`$stdDevPop`~
- ~`$stdDevSamp`~
- ~`$sum`~
- ~`$top`~
- ~`$topN`~

I moved some operators from Expression to Accumulator.


tests/Builder/Accumulator/DenseRankAccumulatorTest.php
```diff
@@ -0,0 +1,55 @@
...
+        $pipeline = new Pipeline(
+            Stage::setWindowFields(
+                partitionBy: Expression::stringFieldPath('state'),
+                sortBy: object(
+                    orderDate: 1,
+                ),
+                output: object(
+                    denseRankOrderDateForState: Accumulator::outputWindow(
+                        Accumulator::denseRank(),
+                    ),
```
GromNaN, line 31: The `outputWindow` is optional when no window property is set. I omitted it in the next test.

tests/Builder/Accumulator/ExpMovingAvgAccumulatorTest.php
```diff
@@ -0,0 +1,59 @@
...
+        $pipeline = new Pipeline(
+            Stage::setWindowFields(
+                partitionBy: Expression::stringFieldPath('stock'),
+                sortBy: object(
+                    date: 1,
+                ),
+                output: object(
+                    expMovingAvgForStock: Accumulator::expMovingAvg(
+                        input: Expression::numberFieldPath('price'),
+                        N: 2,
```
alcaeus, line 51: I noted that `$expMovingAvg` indeed uses an uppercase single-letter parameter name. I'm sure that won't be confusing for people. No need to do something about this now, but I can see ourselves adding some kind of fallback for folks that aren't used to this and use `n` out of habit.

alcaeus (approved):

tests/Builder/Accumulator/ExpMovingAvgAccumulatorTest.php
GromNaN: PHP is case-insensible for method names but not for property or argument names. Static analysis and PHP error will help:
> Fatal error: Uncaught Error: Unknown named parameter $n

alcaeus: Yep, that will indeed help :D

tests/Builder/Accumulator/DenseRankAccumulatorTest.php
GromNaN: Added this note as a comment in the code.

PR #52, opened 2024-01-24
PHPLIB-1342 Add tests on Arithmetic Expression Operators

GromNaN:
Fix [PHPLIB-1342](https://jira.mongodb.org/browse/PHPLIB-1342)

https://www.mongodb.com/docs/manual/reference/operator/aggregation/#arithmetic-expression-operators

- `$abs`
- ~`$add`~
- `$ceil`
- `$divide`
- `$exp`
- `$floor`
- `$ln`
- `$log`
- `$log10`
- `$mod`
- `$multiply`
- `$pow`
- `$round`
- `$sqrt`
- ~`$subtract`~
- `$trunc`

Nothing special

### DIFF
--- a/generator/config/expression/pow.yaml
+++ b/generator/config/expression/pow.yaml
@@ -15,3 +15,17 @@ arguments:
         name: exponent
         type:
             - resolvesToNumber
+tests:
+    -
+        name: 'Example'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/pow/#example'
+        pipeline:
+            -
+                $project:
+                    variance:
+                        $pow:
+                            -
+                                # The builder renders $stdDevPop with the array form, even with a single value
+                                # $stdDevPop: '$scores.score'
+                                $stdDevPop: ['$scores.score']
+                            - 2
--- a/generator/config/expression/abs.yaml
+++ b/generator/config/expression/abs.yaml
@@ -11,3 +11,15 @@ arguments:
         name: value
         type:
             - resolvesToNumber
+tests:
+    -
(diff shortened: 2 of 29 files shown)


generator/config/expression/pow.yaml
```diff
@@ -15,3 +15,16 @@ arguments:
...
+    -
+        name: 'Example'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/pow/#example'
+        pipeline:
+            -
+                $project:
+                    variance:
+                        $pow:
+                            -
+                                # $stdDevPop: '$scores.score'
```
GromNaN, line 29: I've checked the result is the same if the value is in an array or a single value

jmikola: Is this because StdDevPopOperator encodes as a single element, which utilizes its only `$expression` property (a list)? I suppose `$expressions` would be a more descriptive name for the property but I realize the class is auto-generated and derived from the name of the variadic arg itself.

In any event, I think it makes sense to capture this explanation in a YAML comment here to explain exactly why you're changing the expected pipeline.

IIRC, this kind of thing came up in previous PRs, since there are other operators that omit logic to encode using the "short" form.

jmikola (approved):
LGTM with comment explaining why `$stdDevPop` example was changed.

generator/config/expression/pow.yaml
GromNaN: The argument is variadic so we always get an array. That's one reason why I made an accumulator operator distrinct from the expression operator.
Comment added.

PR #53, opened 2024-01-24
PHPLIB-1341 Add tests on Miscellaneous Query Operators

GromNaN:
Fix [PHPLIB-1341](https://jira.mongodb.org/browse/PHPLIB-1341)

https://www.mongodb.com/docs/manual/reference/operator/query/#miscellaneous-operators

- `$comment`
- `$rand`

### DIFF
--- a/generator/config/query/comment.yaml
+++ b/generator/config/query/comment.yaml
@@ -11,3 +11,21 @@ arguments:
         name: comment
         type:
             - string
+tests:
+    -
+        name: 'Attach a Comment to an Aggregation Expression'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/comment/#attach-a-comment-to-an-aggregation-expression'
+        pipeline:
+            -
+                $match:
+                    x:
+                        $gt: 0
+                    $comment: 'Don''t allow negative inputs.'
+            -
+                $group:
+                    _id:
+                        $mod:
+                            - '$x'
+                            - 2
+                    total:
+                        $sum: '$x'
--- a/generator/js2yaml.html
+++ b/generator/js2yaml.html
@@ -124,7 +124,7 @@
             case 'boolean':
                 return object ? ' true' : ' false';
             case 'string':
-                return " '" + object.replace(/'/g, "\\'") + "'";
+                return " '" + object.replace(/'/g, "''") + "'";
             case 'number':
                 return ' ' + object.toString();
             case 'object':
--- a/tests/Builder/Query/CommentOperatorTest.php
+++ b/tests/Builder/Query/CommentOperatorTest.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Tests\Builder\Query;
+
+use MongoDB\Builder\Accumulator;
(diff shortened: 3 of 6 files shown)


tests/Builder/Query/CommentOperatorTest.php
```diff
@@ -0,0 +1,39 @@
...
+ * Test $comment query
+ */
+class CommentOperatorTest extends PipelineTestCase
+{
+    public function testAttachACommentToAnAggregationExpression(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::match(
+                Query::comment('Don\'t allow negative inputs.'),
+                x: Query::gt(0),
```
GromNaN, line 24: Named arguments must be after positional arguments, even when we use variadic.

generator/js2yaml.html
```diff
@@ -124,7 +124,7 @@ <h1>Convert JS examples into Yaml</h1>
             case 'boolean':
                 return object ? ' true' : ' false';
             case 'string':
-                return " '" + object.replace(/'/g, "\\'") + "'";
+                return " '" + object.replace(/'/g, "''") + "'";
```
jmikola, line 127: Is this necessary for the single-quoted style being used above?

generator/config/query/comment.yaml
```diff
@@ -11,3 +11,21 @@ arguments:
...
+tests:
+    -
+        name: 'Attach a Comment to an Aggregation Expression'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/query/comment/#attach-a-comment-to-an-aggregation-expression'
+        pipeline:
+            -
+                $match:
+                    x:
+                        $gt: 0
+                    $comment: 'Don''t allow negative inputs.'
```
jmikola, line 23: Noted that this syntax is specific to [Single-Quoted Style](https://yaml.org/spec/1.2.2/#single-quoted-style). Would it make more sense to use a [Double-Quoted Style](https://yaml.org/spec/1.2.2/#731-double-quoted-style) and avoid escaping altogether, or did you prefer single-quoted for consistency?

tests/Builder/Query/RandOperatorTest.php
```diff
@@ -0,0 +1,39 @@
...
+class RandOperatorTest extends PipelineTestCase
+{
+    public function testSelectRandomItemsFromACollection(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::match(
+                Query::expr(
+                    Expression::lt(
+                        0.5,
+                        Expression::rand(),
```
jmikola, line 25: Since `RandOperator::ENCODE` is `Encode::Object` and the class has no properties, we can be sure that this encodes as an empty BSON document, correct?

jmikola (approved):

tests/Builder/Query/RandOperatorTest.php
GromNaN: Exactly, that's the trick to getting an empty object.

generator/config/query/comment.yaml
GromNaN: I use single quotes everywhere to avoid risky parsing.

PR #54, opened 2024-01-24
PHPLIB-1340 Remove Projection Operators

GromNaN:
Fix [PHPLIB-1340](https://jira.mongodb.org/browse/PHPLIB-1340)

The projection operators can only be used with the `find` command.

https://www.mongodb.com/docs/manual/reference/operator/query/#projection-operators

- ~`$`~ no operator
- ~`$elemMatch`~
- ~`$meta`~ covered by https://github.com/mongodb/mongo-php-builder/pull/38
- ~`$slice`~

Also removed `$filter` from Projection operators, as it's an Expression operator.


generator/config/projection/filter.yaml
```diff
@@ -1,34 +0,0 @@
-# $schema: ../schema.json
```
jmikola, line 1: Why was this file deleted?

generator/config/projection/slice.yaml
```diff
@@ -15,3 +15,41 @@ arguments:
...
+    -
+        name: 'Return an Array with Its First 3 Elements'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/projection/slice/#return-an-array-with-its-first-3-elements'
+        pipeline:
+            -
+                $project:
+                    comments:
+                        # Example uses the short form, the builder always generated the verbose form
+                        # $slice: 3
+                        $slice: [3]
```
jmikola: This doesn't seem correct, as the docs indicate that the array form requires both elements:

> Specifies the number of elements to return in the <arrayField> after skipping the specified number of elements starting from the first element. You must specify both elements.

Also, the short form is `$slice: <number to return>`. When an array form is used, the first value specifies the number to skip and `<number to return>` is the second value.

I would hope the server would just error on this since it's ambiguous, although I could see a case where <number to return> is accepted and allows `$slice` to skip a few elements and return everything else. The server docs don't suggest that's possible, so I assume you can only accomplish that by specifying a very large `<number to return>` and hoping it exceeds the array size.

tests/Builder/BuilderEncoderTest.php
```diff
@@ -157,7 +156,7 @@ public function testExpressionFilter(array $limit, array $expectedLimit): void
     {
         $pipeline = new Pipeline(
             Stage::project(
-                items: Projection::filter(
+                items: Expression::filter(
```
jmikola, line 159: I assume this is related to deleting `filter.yaml`.

generator/config/projection/slice.yaml
```diff
@@ -15,3 +15,41 @@ arguments:
         name: skip
         type:
             - int
+        optional: true
+tests:
```
jmikola: > Also removed `$slice` from Projection operators, as it's an Expression operator.

How is that? `$slice` exists both as a projection and expression operator (each has different APIs):

 * https://www.mongodb.com/docs/manual/reference/operator/aggregation/slice/#mongodb-expression-exp.-slice
 * https://www.mongodb.com/docs/manual/reference/operator/projection/slice/

This PR is dealing with the projection variant.

tests/Builder/Projection/SliceOperatorTest.php
```diff
@@ -0,0 +1,60 @@
...
+        );
+
+        $this->assertSamePipeline(Pipelines::SliceReturnAnArrayWith3ElementsAfterSkippingTheFirstElement, $pipeline);
+    }
+
+    public function testReturnAnArrayWith3ElementsAfterSkippingTheLastElement(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::project(
+                comments: Projection::slice(-1, 3),
```
jmikola: I think the argument order for `Projection::slice` will need to change, and I can see this being confusing for users since it will differ from the server API.

Both the short and long forms accept a `<number to return>` value, and the long form also accepts `<number to skip>`. If the PHP API is going to provide a consistent API without using named parameters, `<number to skip>` should be an optional second parameter.

jmikola (commented):
I see you crossed out the [positional operator ($)](https://www.mongodb.com/docs/manual/reference/operator/projection/positional/). Is there no work to be done for that?

GromNaN:
> I see you crossed out the [positional operator ($)](https://www.mongodb.com/docs/manual/reference/operator/projection/positional/). Is there no work to be done for that?

Maybe something similar to fieldpath then?

tests/Builder/BuilderEncoderTest.php
GromNaN: Yes it is.

generator/config/projection/slice.yaml
GromNaN: Sorry, I meant `$filter`, not `$slice`.

generator/config/projection/slice.yaml
GromNaN: Looks like we will need a custom encoder. 

`$slice: [3]` returns a server error.

> [MongoServerError:](https://github.com/mongodb/mongo-php-builder/pull/54#) Invalid $slice syntax. The given syntax { $slice: [ 3 ] } did not match the find() syntax because :: Location31272: $slice array argument should be of form [skip, limit] :: The given syntax did not match the expression $slice syntax. :: caused by :: Expression $slice takes at least 2 arguments, and at most 3, but 1 were passed in.

Note: the version with 3 arguments is not documented.

jmikola: > the version with 3 arguments is not documented

I apologize for the headache, but that sounds like it warrants a DOCSP ticket. And make sure to tell them you shouldn't be used as the technical reviewer when/if it actually gets implemented (this frequently happens to me when I report server docs issues). It should really be someone from server, since they'll have to confirm what it actually does.

And if it should remain undocumented, server should consider revising this error message.

jmikola (approved):
Since the positional operator can occur anywhere within a field path, I'm not certain a separate builder method is necessary unless you intend to enforce some typing with it and assert that it's well formed (such as ".$" at the end of a path or ".$." within a path).

I'm not convinced that's needed, though.

GromNaN:
I decided to remove all projection operators. They can't be used in `$project` stage as I was expecting. They are dedicated to the `find` command. The `$slice` must be used as an expression (with the array as 1st parameter) and `$elemMatch` is replaced by `$filter`.

generator/config/projection/slice.yaml
GromNaN: https://jira.mongodb.org/browse/DOCS-16617

jmikola (approved):

alcaeus (approved):

PR #56, opened 2024-01-24
PHPLIB-1374 Add tests on all stages

GromNaN:
Fix [PHPLIB-1374](https://jira.mongodb.org/browse/PHPLIB-1374)

https://www.mongodb.com/docs/manual/reference/operator/aggregation-pipeline/

- ~`$addFields`~ already done
- `$bucket`
- `$bucketAuto`
- `$changeStream`
- `$changeStreamSplitLargeEvent`
- `$collStats`
- `$count`
- `$densify`
- `$documents`
- `$facet`
- `$fill`
- `$geoNear`
- `$graphLookup`
- `$group`
- `$indexStats`
- `$limit`
- `$listSampledQueries`
- `$listSearchIndexes`
- `$listSessions`
- `$lookup`
- `$match`
- `$merge`
- `$out`
- `$planCacheStats`
- `$project`
- `$redact`
- `$replaceRoot`
- `$replaceWith`
- `$sample`
- ~`$search`~ (no example)
- ~`$searchMeta`~ (no example)
- `$set`
- `$setWindowFields`
- `$skip`
- `$sort`
- `$sortByCount`
- `$unionWith`
- `$unset`
- `$unwind`

### DIFF
--- a/generator/config/expression/sum.yaml
+++ b/generator/config/expression/sum.yaml
@@ -12,6 +12,7 @@ arguments:
         name: expression
         type:
             - resolvesToNumber
+            - resolvesToArray
         variadic: array
 tests:
     -
--- a/generator/config/expression/sum.yaml
+++ b/generator/config/expression/sum.yaml
@@ -22,6 +22,7 @@ tests:
             -
                 $project:
                     quizTotal:
+                        # Example uses the short form, the builder always generates the verbose form
                         # $sum: '$quizzes'
                         $sum:
                             - '$quizzes'
--- a/generator/config/stage/bucket.yaml
+++ b/generator/config/stage/bucket.yaml
@@ -40,3 +40,62 @@ arguments:
             A document that specifies the fields to include in the output documents in addition to the _id field. To specify the field to include, you must use accumulator expressions.
             If you do not specify an output document, the operation returns a count field containing the number of documents in each bucket.
             If you specify an output document, only the fields specified in the document are returned; i.e. the count field is not returned unless it is explicitly included in the output document.
+tests:
+    -
+        name: 'Bucket by Year and Filter by Bucket Results'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/bucket/#bucket-by-year-and-filter-by-bucket-results'
+        pipeline:
+            -
+                $bucket:
+                    groupBy: '$year_born'
+                    boundaries: [1840, 1850, 1860, 1870, 1880]
+                    default: 'Other'
+                    output:
+                        count:
+                            $sum: 1
+                        artists:
+                            $push:
+                                name:
+                                    $concat:
+                                        - '$first_name'
+                                        - ' '
+                                        - '$last_name'
+                                year_born: '$year_born'
+            -
+                $match:
+                    count:
+                        $gt: 3
+    -
+        name: 'Use $bucket with $facet to Bucket by Multiple Fields'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/bucket/#use--bucket-with--facet-to-bucket-by-multiple-fields'
+        pipeline:
+            -
+                $facet:
+                    price:
+                        -
+                            $bucket:
+                                groupBy: '$price'
+                                boundaries: [0, 200, 400]
+                                default: 'Other'
+                                output:
+                                    count:
+                                        $sum: 1
+                                    artwork:
+                                        $push:
+                                            title: '$title'
+                                            price: '$price'
+                                    averagePrice:
+                                        $avg: '$price'
+                    year:
+                        -
+                            $bucket:
+                                groupBy: '$year'
+                                boundaries: [1890, 1910, 1920, 1940]
+                                default: 'Unknown'
+                                output:
+                                    count:
+                                        $sum: 1
+                                    artwork:
+                                        $push:
+                                            title: '$title'
+                                            year: '$year'
--- a/generator/config/stage/collStats.yaml
+++ b/generator/config/stage/collStats.yaml
@@ -3,14 +3,30 @@ name: $collStats
 link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/collStats/'
 type:
     - stage
-encode: single
+encode: object
 description: |
     Returns statistics regarding a collection or view.
 arguments:
     -
-        name: config
+        name: latencyStats
         type:
             - object
+        optional: true
+    -
+        name: storageStats
+        type:
+            - object
+        optional: true
+    -
+        name: count
+        type:
+            - object # empty object
+        optional: true
+    -
+        name: queryExecStats
+        type:
+            - object # empty object
+        optional: true
 tests:
     -
         name: 'latencyStats Document'
--- a/generator/config/stage/collStats.yaml
+++ b/generator/config/stage/collStats.yaml
@@ -11,3 +11,33 @@ arguments:
         name: config
         type:
             - object
+tests:
+    -
+        name: 'latencyStats Document'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/collStats/#latencystats-document'
+        pipeline:
+            -
+                $collStats:
+                    latencyStats:
+                        histograms: true
+    -
+        name: 'storageStats Document'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/collStats/#storagestats-document'
+        pipeline:
+            -
+                $collStats:
+                    storageStats: {}
+    -
+        name: 'count Field'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/collStats/#count-field'
+        pipeline:
+            -
+                $collStats:
+                    count: {}
+    -
+        name: 'queryExecStats Document'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/collStats/#queryexecstats-document'
+        pipeline:
+            -
+                $collStats:
+                    queryExecStats: {}
--- a/generator/config/stage/currentOp.yaml
+++ b/generator/config/stage/currentOp.yaml
@@ -6,3 +6,54 @@ type:
 encode: object
 description: |
     Returns information on active and/or dormant operations for the MongoDB deployment. To run, use the db.aggregate() method.
+arguments:
+    -
+        name: allUsers
+        type:
+            - bool
+        optional: true
+    -
+        name: idleConnections
+        type:
+            - bool
+        optional: true
+    -
+        name: idleCursors
+        type:
+            - bool
+        optional: true
+    -
+        name: idleSessions
+        type:
+            - bool
+        optional: true
+    -
+        name: localOps
+        type:
+            - bool
+        optional: true
+tests:
+    -
+        name: 'Inactive Sessions'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/currentOp/#inactive-sessions'
+        pipeline:
+            -
+                $currentOp:
+                    allUsers: true
+                    idleSessions: true
+            -
+                $match:
+                    active: false
+                    transaction:
+                        $exists: true
+    -
+        name: 'Sampled Queries'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/currentOp/#sampled-queries'
+        pipeline:
+            -
+                $currentOp:
+                    allUsers: true
+                    localOps: true
+            -
+                $match:
+                    desc: 'query analyzer'
--- a/generator/config/stage/densify.yaml
+++ b/generator/config/stage/densify.yaml
@@ -28,3 +28,29 @@ arguments:
             - object # Range
         description: |
             Specification for range based densification.
+tests:
+    -
+        name: 'Densify Time Series Data'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/densify/#densify-time-series-data'
+        pipeline:
+            -
+                $densify:
+                    field: 'timestamp'
+                    range:
+                        step: 1
+                        unit: 'hour'
+                        bounds:
+                            - !bson_utcdatetime '2021-05-18T00:00:00.000Z'
+                            - !bson_utcdatetime '2021-05-18T08:00:00.000Z'
+    -
+        name: 'Densifiction with Partitions'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/densify/#densifiction-with-partitions'
+        pipeline:
+            -
+                $densify:
+                    field: 'altitude'
+                    partitionByFields:
+                        - 'variety'
+                    range:
+                        bounds: 'full'
+                        step: 200
--- a/generator/config/stage/facet.yaml
+++ b/generator/config/stage/facet.yaml
@@ -21,6 +21,7 @@ tests:
                 $facet:
                     categorizedByTags:
                         -
+                            # The builder uses the verbose form of the $unwind operator
                             # $unwind: '$tags'
                             $unwind:
                                 path: '$tags'
--- a/generator/config/stage/facet.yaml
+++ b/generator/config/stage/facet.yaml
@@ -12,3 +12,39 @@ arguments:
         type:
             - pipeline
         variadic: object
+tests:
+    -
+        name: 'Example'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/facet/#example'
+        pipeline:
+            -
+                $facet:
+                    categorizedByTags:
+                        -
+                            # $unwind: '$tags'
+                            $unwind:
+                                path: '$tags'
+                        -
+                            $sortByCount: '$tags'
+                    categorizedByPrice:
+                        -
+                            $match:
+                                price:
+                                    # The example uses an int, but the builder requires a bool
+                                    # $exists: 1
+                                    $exists: true
+                        -
+                            $bucket:
+                                groupBy: '$price'
+                                boundaries: [0, 150, 200, 300, 400]
+                                default: 'Other'
+                                output:
+                                    count:
+                                        $sum: 1
+                                    titles:
+                                        $push: '$title'
+                    categorizedByYears(Auto):
+                        -
+                            $bucketAuto:
+                                groupBy: '$year'
+                                buckets: 4
--- a/generator/config/stage/fill.yaml
+++ b/generator/config/stage/fill.yaml
@@ -40,3 +40,71 @@ arguments:
         description: |
             Specifies an object containing each field for which to fill missing values. You can specify multiple fields in the output object.
             The object name is the name of the field to fill. The object value specifies how the field is filled.
+tests:
+    -
+        name: 'Fill Missing Field Values with a Constant Value'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/fill/#fill-missing-field-values-with-a-constant-value'
+        pipeline:
+            -
+                $fill:
+                    output:
+                        bootsSold:
+                            value: 0
+                        sandalsSold:
+                            value: 0
+                        sneakersSold:
+                            value: 0
+    -
+        name: 'Fill Missing Field Values with Linear Interpolation'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/fill/#fill-missing-field-values-with-linear-interpolation'
+        pipeline:
+            -
+                $fill:
+                    sortBy:
+                        time: 1
+                    output:
+                        price:
+                            method: 'linear'
+    -
+        name: 'Fill Missing Field Values Based on the Last Observed Value'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/fill/#fill-missing-field-values-based-on-the-last-observed-value'
+        pipeline:
+            -
+                $fill:
+                    sortBy:
+                        date: 1
+                    output:
+                        score:
+                            method: 'locf'
+    -
+        name: 'Fill Data for Distinct Partitions'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/fill/#fill-data-for-distinct-partitions'
+        pipeline:
+            -
+                $fill:
+                    sortBy:
+                        date: 1
+                    partitionBy:
+                        restaurant: '$restaurant'
+                    output:
+                        score:
+                            method: 'locf'
+    -
+        name: 'Indicate if a Field was Populated Using $fill'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/fill/#indicate-if-a-field-was-populated-using--fill'
+        pipeline:
+            -
+                $set:
+                    valueExisted:
+                        $ifNull:
+                            -
+                                $toBool:
+                                    $toString: '$score'
+                            - false
+            -
+                $fill:
+                    sortBy:
+                        date: 1
+                    output:
+                        score:
+                            method: 'locf'
--- a/generator/config/stage/graphLookup.yaml
+++ b/generator/config/stage/graphLookup.yaml
@@ -60,3 +60,49 @@ arguments:
         optional: true
         description: |
             A document specifying additional conditions for the recursive search. The syntax is identical to query filter syntax.
+tests:
+    -
+        name: 'Within a Single Collection'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/graphLookup/#within-a-single-collection'
+        pipeline:
+            -
+                $graphLookup:
+                    from: 'employees'
+                    startWith: '$reportsTo'
+                    connectFromField: 'reportsTo'
+                    connectToField: 'name'
+                    as: 'reportingHierarchy'
+    -
+        name: 'Across Multiple Collections'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/graphLookup/#across-multiple-collections'
+        pipeline:
+            -
+                $graphLookup:
+                    from: 'airports'
+                    startWith: '$nearestAirport'
+                    connectFromField: 'connects'
+                    connectToField: 'airport'
+                    maxDepth: 2
+                    depthField: 'numConnections'
+                    as: 'destinations'
+    -
+        name: 'With a Query Filter'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/graphLookup/#with-a-query-filter'
+        pipeline:
+            -
+                $match:
+                    name: 'Tanya Jordan'
+            -
+                $graphLookup:
+                    from: 'people'
+                    startWith: '$friends'
+                    connectFromField: 'friends'
+                    connectToField: 'name'
+                    as: 'golfers'
+                    restrictSearchWithMatch:
+                        hobbies: 'golf'
+            -
+                $project:
+                    name: 1
+                    friends: 1
+                    connections who play golf: '$golfers.name'
--- a/generator/config/stage/group.yaml
+++ b/generator/config/stage/group.yaml
@@ -18,5 +18,105 @@ arguments:
         type:
             - accumulator
         variadic: object
+        variadicMin: 0
         description: |
             Computed using the accumulator operators.
+tests:
+    -
+        name: 'Count the Number of Documents in a Collection'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/group/#count-the-number-of-documents-in-a-collection'
+        pipeline:
+            -
+                $group:
+                    _id: ~
+                    count:
+                        $count: {}
+    -
+        name: 'Retrieve Distinct Values'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/group/#retrieve-distinct-values'
+        pipeline:
+            -
+                $group:
+                    _id: '$item'
+    -
+        name: 'Group by Item Having'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/group/#group-by-item-having'
+        pipeline:
+            -
+                $group:
+                    _id: '$item'
+                    totalSaleAmount:
+                        $sum:
+                            $multiply:
+                                - '$price'
+                                - '$quantity'
+            -
+                $match:
+                    totalSaleAmount:
+                        $gte: 100
+    -
+        name: 'Calculate Count Sum and Average'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/group/#calculate-count--sum--and-average'
+        pipeline:
+            -
+                $match:
+                    date:
+                        $gte: !bson_utcdatetime '2014-01-01'
+                        $lt: !bson_utcdatetime '2015-01-01'
+            -
+                $group:
+                    _id:
+                        $dateToString:
+                            format: '%Y-%m-%d'
+                            date: '$date'
+                    totalSaleAmount:
+                        $sum:
+                            $multiply:
+                                - '$price'
+                                - '$quantity'
+                    averageQuantity:
+                        $avg: '$quantity'
+                    count:
+                        $sum: 1
+            -
+                $sort:
+                    totalSaleAmount: -1
+    -
+        name: 'Group by null'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/group/#group-by-null'
+        pipeline:
+            -
+                $group:
+                    _id: ~
+                    totalSaleAmount:
+                        $sum:
+                            $multiply:
+                                - '$price'
+                                - '$quantity'
+                    averageQuantity:
+                        $avg: '$quantity'
+                    count:
+                        $sum: 1
+    -
+        name: 'Pivot Data'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/group/#pivot-data'
+        pipeline:
+            -
+                $group:
+                    _id: '$author'
+                    books:
+                        $push: '$title'
+    -
+        name: 'Group Documents by author'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/group/#group-documents-by-author'
+        pipeline:
+            -
+                $group:
+                    _id: '$author'
+                    books:
+                        $push: '$$ROOT'
+            -
+                $addFields:
+                    totalCopies:
+                        # $sum: '$books.copies'
+                        $sum: ['$books.copies']
--- a/generator/config/stage/merge.yaml
+++ b/generator/config/stage/merge.yaml
@@ -12,7 +12,7 @@ arguments:
         name: into
         type:
             - string
-            #- OutCollection
+            - object # OutCollection
         description: |
             The output collection.
     -
@@ -34,6 +34,7 @@ arguments:
         name: whenMatched
         type:
             - string # WhenMatched
+            - pipeline
         optional: true
         description: |
             The behavior of $merge if a result document and an existing document in the collection have the same value for the specified on field(s).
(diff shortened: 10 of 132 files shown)


generator/config/stage/currentOp.yaml
```diff
@@ -6,3 +6,54 @@ type:
...
+    -
+        name: idleSessions
+        type:
+            - bool
+        optional: true
+    -
+        name: localOps
+        type:
+            - bool
+        optional: true
```
GromNaN, line 34: Parameters were missing, I added them as they are only booleans.

generator/config/stage/collStats.yaml
```diff
@@ -11,3 +11,33 @@ arguments:
...
             - object
+tests:
+    -
+        name: 'latencyStats Document'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/collStats/#latencystats-document'
+        pipeline:
+            -
+                $collStats:
+                    latencyStats:
+                        histograms: true
```
GromNaN, line 38: The config contains nested object definition. That's why I kept a generic "config" object argument. DX could be improved with advanced type in phpdoc.

generator/config/stage/densify.yaml
```diff
@@ -28,3 +28,29 @@ arguments:
...
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/densify/#densifiction-with-partitions'
+        pipeline:
+            -
+                $densify:
+                    field: 'altitude'
+                    partitionByFields:
+                        - 'variety'
+                    range:
+                        bounds: 'full'
+                        step: 200
```
GromNaN, line 56: The "range" config could have a factory and use and enum for "bounds".

generator/config/stage/densify.yaml
```diff
@@ -28,3 +28,29 @@ arguments:
...
+    -
+        name: 'Densify Time Series Data'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/densify/#densify-time-series-data'
+        pipeline:
+            -
+                $densify:
+                    field: 'timestamp'
+                    range:
+                        step: 1
+                        unit: 'hour'
```
GromNaN, line 41: We need to create an enum for time unit, used in a lot of places.

generator/config/stage/fill.yaml
```diff
@@ -40,3 +40,71 @@ arguments:
...
+        name: 'Fill Missing Field Values with Linear Interpolation'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/fill/#fill-missing-field-values-with-linear-interpolation'
+        pipeline:
+            -
+                $fill:
+                    sortBy:
+                        time: 1
+                    output:
+                        price:
+                            method: 'linear'
```
GromNaN, line 67: We an create a factory for the output subtype.

generator/config/stage/group.yaml
```diff
@@ -18,5 +18,105 @@ arguments:
...
+                    _id: ~
+                    count:
+                        $count: {}
+    -
+        name: 'Retrieve Distinct Values'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/group/#retrieve-distinct-values'
+        pipeline:
+            -
+                $group:
+                    _id: '$item'
```
GromNaN, line 40: This example shows that an accumulator is not required. I reduced `variadicMin` to 0.

generator/config/stage/merge.yaml
```diff
@@ -44,3 +45,136 @@ arguments:
...
+                    salaries:
+                        $sum: '$salary'
+            -
+                $merge:
+                    into:
+                        db: 'reporting'
+                        coll: 'budgets'
+                    on: '_id'
+                    whenMatched: 'replace'
+                    whenNotMatched: 'insert'
```
GromNaN, line 67: 2 enums should be created for [this values](https://www.mongodb.com/docs/manual/reference/operator/aggregation/merge/#syntax): `WhenMatched<replace|keepExisting|merge|fail>` (+ Pipeline allowed) and `WhenNotMatched<insert|discard|fail>`

generator/config/stage/out.yaml
```diff
@@ -3,26 +3,38 @@ name: $out
 link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/out/'
 type:
     - stage
-encode: object
+encode: single
```
GromNaN, line 6: I have to change this type, because an object with only `{ coll }` is not accepted. The value must be a string or an object with `{ coll, db }`

We could have a factory that return a string or an object, that could be used in `$merge` also.

generator/config/stage/set.yaml
```diff
@@ -13,3 +13,61 @@ arguments:
...
+        pipeline:
+            -
+                $set:
+                    totalHomework:
+                        # The $sum expression is always build as an array, even if the value is an array field name
+                        # $sum: '$homework'
+                        $sum: ['$homework']
+                    totalQuiz:
+                        # $sum: '$quiz'
+                        $sum: ['$quiz']
```
GromNaN, line 29: I've checked all this variations from the doc examples.

generator/config/stage/setWindowFields.yaml
```diff
@@ -28,3 +28,117 @@ arguments:
...
+            -
+                $setWindowFields:
+                    partitionBy: '$state'
+                    sortBy:
+                        orderDate: 1
+                    output:
+                        cumulativeQuantityForState:
+                            $sum: '$quantity'
+                            window:
+                                documents: ['unbounded', 'current']
```
GromNaN, line 45: We need an enum for `unbound` / `current`...

tests/Builder/Stage/UnionWithStageTest.php
```diff
@@ -0,0 +1,77 @@
...
+                fn ($year) => Stage::unionWith(
+                    coll: 'sales_' . $year,
+                    pipeline: new Pipeline(
+                        Stage::set(
+                            _id: (string) $year,
+                        ),
+                    ),
+                ),
+                range(2018, 2020),
+            ),
```
GromNaN: Experimentation to use `array_map`. But array unpacking can't be used in the middle of 2 other positional arguments. So I had to unpack the result of an `array_merge`.

tests/Builder/Stage/ProjectStageTest.php
```diff
@@ -0,0 +1,141 @@
...
+        $pipeline = new Pipeline(
+            Stage::project(
+                myArray: [Expression::fieldPath('x'), Expression::fieldPath('y')],
+            ),
+        );
+
+        $this->assertSamePipeline(Pipelines::ProjectProjectNewArrayFields, $pipeline);
+    }
+
+    public function testSuppress_idFieldInTheOutputDocuments(): void
```
GromNaN: I need to fix the generator for this error.

generator/config/stage/out.yaml
```diff
@@ -3,26 +3,38 @@ name: $out
...
         description: |
             Target database name to write documents from $out to.
+tests:
     -
-        name: timeseries
-        type:
-            - object
-        optional: true
-        description: |
-            If set, the aggregation stage will use these options to create or replace a time-series collection in the given namespace.
```
GromNaN, line 28: I removed this argument, since it needs to be inside the `coll` object. But I could not find an example where it is used. I don't remember where I found it.

generator/config/stage/bucket.yaml
```diff
@@ -40,3 +40,62 @@ arguments:
...
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/bucket/#bucket-by-year-and-filter-by-bucket-results'
+        pipeline:
+            -
+                $bucket:
+                    groupBy: '$year_born'
+                    boundaries: [1840, 1850, 1860, 1870, 1880]
+                    default: 'Other'
+                    output:
+                        count:
+                            $sum: 1
```
GromNaN, line 55: Most examples use: `$sum: 1` where `$count: {}` could be used.

tests/Builder/Stage/CurrentOpStageTest.php
```diff
@@ -0,0 +1,47 @@
...
+    public function testInactiveSessions(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::currentOp(
+                allUsers: true,
+                idleSessions: true,
+            ),
+            Stage::match(
+                active: false,
+                transaction: Query::exists(true),
```
GromNaN, line 26: We could have `true` as default value. So that it can be written: `Query::exists(),`

tests/Builder/Stage/SortStageTest.php
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Test $sort stage
+ */
+class SortStageTest extends PipelineTestCase
+{
+    public function testAscendingDescendingSort(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::sort(
+                object(
```
GromNaN, line 23: The `$sort` operator will need some changes to be variadic object [PHPLIB-1269](https://jira.mongodb.org/browse/PHPLIB-1269). But this sort specs are shared with other operators as property. So we also need a SortSpec factory that returns an object.

generator/config/stage/collStats.yaml
alcaeus: SGTM. I don't think this stage will be used very often.

generator/config/stage/bucket.yaml
alcaeus: I believe this is left over from times before `$count` was introduced.

tests/Builder/Stage/CurrentOpStageTest.php
alcaeus: Sounds like a good idea 👍

tests/Builder/Stage/SortStageTest.php
alcaeus: I noticed that in a number of other examples. Ideally, we would be able to write this like so:
```php
Stage::sort(
    age: -1,
    posts: 1,
);
```

IIRC there were some other instances where we pass sort specifications, I'll dig into some pipelines to find those occurrences to see if this applies there as well.

generator/config/expression/sum.yaml
```diff
@@ -12,6 +12,7 @@ arguments:
         name: expression
         type:
             - resolvesToNumber
+            - resolvesToArray
```
alcaeus, line 15: I assume this is implicitly tested through some of the examples for stages, which is also what necessitated this change?

tests/Builder/Stage/UnionWithStageTest.php
GromNaN: The `Pipeline` class could expand lists of provided as argument.

```php
new Pipeline(
    Stage::set(/* ... */),
    array_map(/* ... */),
    Stage::sort(/* ... */),
);
```

See https://github.com/mongodb/mongo-php-builder/pull/57

tests/Builder/Stage/FacetStageTest.php
```diff
@@ -0,0 +1,62 @@
...
+ * Test $facet stage
+ */
+class FacetStageTest extends PipelineTestCase
+{
+    public function testExample(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::facet(
+                ...[
+                    'categorizedByYears(Auto)' => new Pipeline(
```
alcaeus, line 26: Noted that this field is the reason for using the splat operator. I think it's fine to leave this in as an example when using this strategy can be helpful instead of fixing the example.

tests/Builder/Stage/DensifyStageTest.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        $pipeline = new Pipeline(
+            Stage::densify(
+                field: 'timestamp',
+                range: object(
+                    step: 1,
+                    unit: 'hour',
+                    bounds: [
+                        new UTCDateTime(new DateTimeImmutable('2021-05-18T00:00:00.000Z')),
+                        new UTCDateTime(new DateTimeImmutable('2021-05-18T08:00:00.000Z')),
```
alcaeus, line 46: Noted that `UTCDateTime` does not supporting creating a date from a time string. I created [PHPC-2352](https://jira.mongodb.org/browse/PHPC-2352) to track this.

generator/config/stage/fill.yaml
alcaeus: Agree. This would look better:
```php
Stage::fill(
    sortBy: object(date: 1),
    output: object(
        score: Fill::locf,
        price: Fill::linear,
    )
);
```

tests/Builder/Stage/OutStageTest.php
```diff
@@ -0,0 +1,54 @@
...
+                books: Accumulator::push(
+                    Expression::stringFieldPath('title'),
+                ),
+            ),
+            Stage::out(
+                object(
+                    db: 'reporting',
+                    coll: 'authors',
+                ),
+            ),
```
alcaeus, line 34: I dislike this signature, as passing an object seems overly complex. This is one of those cases where a separate (not auto-generated) encoder would work better, so we can essentially have the following signature:
```php
function out(string $coll, ?string $db = null)
```

tests/Builder/Stage/ProjectStageTest.php
```diff
@@ -0,0 +1,141 @@
...
+        );
+
+        $this->assertSamePipeline(Pipelines::ProjectIncludeSpecificFieldsInOutputDocuments, $pipeline);
+    }
+
+    public function testProjectNewArrayFields(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::project(
+                myArray: [Expression::fieldPath('x'), Expression::fieldPath('y')],
```
alcaeus: Nit:
```suggestion
                myArray: [
                    Expression::fieldPath('x'),
                    Expression::fieldPath('y'),
                ],
```

tests/Builder/Stage/SortStageTest.php
```diff
@@ -0,0 +1,49 @@
...
+
+    public function testTextScoreMetadataSort(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::match(
+                Query::text('operating'),
+            ),
+            Stage::sort(
+                object(
+                    score: ['$meta' => 'textScore'],
```
alcaeus, line 41: Should we have a factory for this? Consider:
```suggestion
                    score: Expression::sortMeta('textScore'),
```
Alternatively, an enum could also work fine here:
```suggestion
                    score: SortMeta::TextScore,
```

tests/Builder/Stage/UnionWithStageTest.php
alcaeus: For the time being, I'd format it manually the way it was done in the second test. I know that's a bit more effort to write, the `...array_merge()` logic with `array_map` is a bit fancy to understand.

tests/Builder/Stage/SortStageTest.php
GromNaN: It will be covered by https://github.com/mongodb/mongo-php-builder/pull/38

tests/Builder/Stage/UnionWithStageTest.php
GromNaN: Ok, reverted.

generator/config/stage/densify.yaml
jmikola: Is this ticketed?

generator/config/stage/fill.yaml
jmikola: @alcaeus: That's basically what you did for ODM already, yes?

generator/config/stage/out.yaml
jmikola: @GromNaN: This reminds me of another operator where you were always encoding the long form (array I think) instead of a conditional short form. Perhaps that was `$type`? Although I'm not sure if it's worth addressing that since it isn't technically required like it is for `$out`.

tests/Builder/Stage/OutStageTest.php
jmikola: Related to https://github.com/mongodb/mongo-php-builder/pull/56#discussion_r1466456462

tests/Builder/Stage/FacetStageTest.php
jmikola: This is also a case where I think a comment explaining the necessity of splat is helpful (even if we repeat it throughout tests whenever this comes up).

generator/config/stage/bucket.yaml
jmikola: > where `$count: {}` could be used.

Just to confirm, you're referring to the [`$count`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/count/) aggregation operator, correct? The syntax is `{ $count: <string> }`, so I wasn't sure if you were just abbreviating or referring to something else.

generator/config/stage/collStats.yaml
jmikola: I agree it will seldom be used, if at all.

But it seems a bit arbitrary to use `config` here, as it isn't anything found in the `$collStats` API. Was it really to much to model the [various options](https://www.mongodb.com/docs/manual/reference/operator/aggregation/collStats/#mongodb-pipeline-pipe.-collStats) like you did with `currentOp.yaml`?

The options themselves don't seem terribly complex, so I don't see a big benefit to not modeling them unless we're concerned about the server API changing.

generator/config/stage/densify.yaml
```diff
@@ -28,3 +28,29 @@ arguments:
...
             Specification for range based densification.
+tests:
+    -
+        name: 'Densify Time Series Data'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/densify/#densify-time-series-data'
+        pipeline:
+            -
+                $densify:
+                    field: 'timestamp'
+                    range:
```
jmikola, line 39: Did you mean to model the `range` option? I see you have `object # Range` above.

tests/Builder/Stage/DocumentsStageTest.php
```diff
@@ -0,0 +1,56 @@
...
+
+        $this->assertSamePipeline(Pipelines::DocumentsTestAPipelineStage, $pipeline);
+    }
+
+    public function testUseADocumentsStageInALookupStage(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::match(),
+            Stage::lookup(
+                localField: 'zip',
```
jmikola, line 42: I realize these are just modeled as string values, but do you think some users might try to use `Expression::fieldPath()` for these?

I assume that would be problematic because the builder would prepend "$", but I'm asking as it seems to be a potential way for users to shoot themselves in the foot.

generator/config/stage/facet.yaml
```diff
@@ -12,3 +12,39 @@ arguments:
...
+tests:
+    -
+        name: 'Example'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/facet/#example'
+        pipeline:
+            -
+                $facet:
+                    categorizedByTags:
+                        -
+                            # $unwind: '$tags'
```
jmikola, line 25: Is this another case where the builder only ever generates the long form? Is there a ticket to revisit that at some point?

Either way, an explanatory comment here explaining why you altered the docs example would be helpful.

tests/Builder/Stage/FacetStageTest.php
```diff
@@ -0,0 +1,62 @@
...
+                    Stage::unwind(
+                        Expression::arrayFieldPath('tags'),
+                    ),
+                    Stage::sortByCount(
+                        Expression::arrayFieldPath('tags'),
+                    ),
+                ),
+                categorizedByPrice: new Pipeline(
+                    Stage::match(
+                        price: Query::exists(true),
```
jmikola, line 43: Potentially replace this with `exists()` if you make `true` the default (per earlier comment).

tests/Builder/Stage/FillStageTest.php
```diff
@@ -0,0 +1,110 @@
...
+ * Test $fill stage
+ */
+class FillStageTest extends PipelineTestCase
+{
+    public function testFillDataForDistinctPartitions(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::fill(
+                sortBy: object(
+                    date: 1,
```
jmikola, line 24: Not sure if this came up elsewhere, but did you plan to make enums for sort ordering like "Asc" and "Desc", or do you think users would prefer using `1` and `-1`, respectively?

I wouldn't push for the enums as it seems overly verbose, but I'd defer to your judgement if you think it'd be helpful for some users.

generator/config/stage/fill.yaml
jmikola: @GromNaN: Will you be creating a ticket to track addition of an `output` factory?

tests/Builder/Stage/GeoNearStageTest.php
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * Test $geoNear stage
+ */
+class GeoNearStageTest extends PipelineTestCase
+{
+    public function testMaximumDistance(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::geoNear(
+                near: object(
```
jmikola, line 24: If a GeoJSON library is implemented, would you plan to clone this test and use a Point object here instead of `object()`? Similar to some of the tests that have enum variations?

tests/Builder/Stage/GeoNearStageTest.php
```diff
@@ -0,0 +1,123 @@
...
+                    type: 'Point',
+                    coordinates: [-73.99279, 40.719296],
+                ),
+                distanceField: 'dist.calculated',
+                maxDistance: 2,
+                query: Query::query(
+                    category: 'Parks',
+                ),
+                includeLocs: 'dist.location',
```
jmikola, line 33: `distanceField` and `includeLocs` seem like other places where users might inadvertently try to use a field path. Would that be outright rejected if the YAML defines the option as just a `string` and not something that resolves to a string?

tests/Builder/Stage/GeoNearStageTest.php
```diff
@@ -0,0 +1,123 @@
...
+        );
+
+        $this->assertSamePipeline(Pipelines::GeoNearSpecifyWhichGeospatialIndexToUse, $pipeline);
+    }
+
+    public function testWithBoundLetOption(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::lookup(
+                from: 'places',
```
jmikola, line 86: Would you want to conditionally model collection and database names? I'm just wondering if integrating libraries might want to hook into this to supply those values (e.g. working with Laravel database prefixes).

Nothing to solve here, but maybe something you want to ticket for future work. In any event, a plain string should also be accepted.

generator/config/stage/graphLookup.yaml
```diff
@@ -60,3 +60,49 @@ arguments:
...
+                    connectFromField: 'friends'
+                    connectToField: 'name'
+                    as: 'golfers'
+                    restrictSearchWithMatch:
+                        hobbies: 'golf'
+            -
+                $project:
+                    name: 1
+                    friends: 1
+                    connections who play golf: '$golfers.name'
```
jmikola, line 108: I think it's ridiculous that the docs use field names like this. It's a missed opportunity to encourage good naming practices (e.g. camelCase, snake_case).

generator/config/stage/group.yaml
```diff
@@ -18,5 +18,105 @@ arguments:
...
+        pipeline:
+            -
+                $group:
+                    _id: '$author'
+                    books:
+                        $push: '$$ROOT'
+            -
+                $addFields:
+                    totalCopies:
+                        # $sum: '$books.copies'
```
jmikola, line 121: Comment explaining why this was changed.

tests/Builder/Stage/ListLocalSessionsStageTest.php
```diff
@@ -0,0 +1,50 @@
...
+        $this->assertSamePipeline(Pipelines::ListLocalSessionsListAllLocalSessionsForTheCurrentUser, $pipeline);
+    }
+
+    public function testListAllLocalSessionsForTheSpecifiedUsers(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::listLocalSessions(
+                users: [
+                    object(user: 'myAppReader', db: 'test'),
+                ],
```
jmikola, line 44: Out of curiosity, is there really any reason to use `object()` here instead of an associative array? I assume both would work the same, yes?

tests/Builder/Stage/ListSearchIndexesStageTest.php
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Test $listSearchIndexes stage
+ */
+class ListSearchIndexesStageTest extends PipelineTestCase
+{
+    public function testReturnASingleSearchIndexById(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::listSearchIndexes(
+                id: '6524096020da840844a4c4a7',
```
jmikola, line 20: This is outside the scope of the PHP builder (I see a string is used in [`$listSearchIndexes`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/listSearchIndexes)), but why the heck is Atlas reporting search indexes as 24-character hexadecimal strings instead of an ObjectId type?

tests/Builder/Stage/LookupStageTest.php
```diff
@@ -0,0 +1,161 @@
...
+                foreignField: 'item',
+                as: 'fromItems',
+            ),
+            Stage::replaceRoot(
+                Expression::mergeObjects(
+                    Expression::arrayElemAt(
+                        Expression::arrayFieldPath('fromItems'),
+                        0,
+                    ),
+                    Expression::variable('ROOT'),
```
jmikola, line 151: Did you create enums for these, or is that still future work in the backlog?

generator/config/stage/merge.yaml
```diff
@@ -12,7 +12,7 @@ arguments:
         name: into
         type:
             - string
-            #- OutCollection
+            - object # OutCollection
```
jmikola, line 15: Is there an outstanding ticket to model OutCollection? This may relate to my earlier comment about modeling or typing collection and database names, since libraries may want to hook into that.

generator/config/stage/merge.yaml
```diff
@@ -34,6 +34,7 @@ arguments:
         name: whenMatched
         type:
             - string # WhenMatched
+            - pipeline
```
jmikola, line 37: The [$merge: Syntax](https://www.mongodb.com/docs/manual/reference/operator/aggregation/merge/#syntax) docs aren't entirely clear here, as they say:

> whenMatched: <replace|keepExisting|merge|fail|pipeline>,  // Optional

I originally thought all of those were supported string values, but "pipeline" actually refers to a pipeline array.

Just to confirm, there's no reason you actually need to enforce the pipeline limitations there, correct? According to the docs, only a few stages are supported in this context. I'm fine relying on the server to enforce that.

generator/config/stage/merge.yaml
jmikola: I was just about to leave a comment on `# WhenMatched` in the option definition above.

Is this todo item ticketed?

tests/Builder/Stage/MergeStageTest.php
```diff
@@ -0,0 +1,188 @@
...
+                dept: Expression::fieldPath('_id.dept'),
+                fiscal_year: Expression::fieldPath('_id.fiscal_year'),
+                employees: 1,
+            ),
+            Stage::merge(
+                into: object(
+                    db: 'reporting',
+                    coll: 'orgArchive',
+                ),
+                on: ['dept', 'fiscal_year'],
```
jmikola, line 123: I'm beating a dead horse here, but this left me wondering if some users might try to use `fieldPath()` methods here. Would that yield an error during static analysis?

generator/config/stage/merge.yaml
```diff
@@ -44,3 +45,136 @@ arguments:
...
+                    whenNotMatched: 'insert'
+    -
+        name: 'Use the Pipeline to Customize the Merge'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/merge/#use-the-pipeline-to-customize-the-merge'
+        pipeline:
+            -
+                $match:
+                    date:
+                        $gte: !bson_utcdatetime 1557187200000
+                        $lt: !bson_utcdatetime 1557273600000
```
jmikola: Why did you opt to use integers here instead of the "2019-05-07 and "2019-05-08" strings from the examples? Likewise for the corresponding PHP test.

I noted other cases where you used prettier date values in the YAML/PHP tests.

jmikola (commented):
Publishing the first round of feedback.

generator/config/stage/fill.yaml
alcaeus: Indeed: https://github.com/doctrine/mongodb-odm/blob/c49d55a8ea5f96c19c0950485e2fd959ffe9efca/tests/Doctrine/ODM/MongoDB/Tests/Aggregation/Stage/FillTest.php#L23..L25

generator/config/stage/bucket.yaml
alcaeus: @jmikola the docs you linked to are for the `$count` aggregation **stage**. MongoDB 5.0 introduced a `$count` aggregation **accumulator** (with the same name to make the confusion perfect) that serves as a shortcut for `{ $sum: 1 }`: [$count (aggregation accumulator)](https://www.mongodb.com/docs/manual/reference/operator/aggregation/count-accumulator/)

generator/config/stage/merge.yaml
alcaeus: I agree, duplicating these limitations here might be quite troublesome, both in duplicating server functionality and in making the type system agree with this.

generator/config/stage/collStats.yaml
GromNaN: Ok, I added the 4 parameters.

generator/config/stage/densify.yaml
GromNaN: Here [PHPLIB-1370](https://jira.mongodb.org/browse/PHPLIB-1370)

generator/config/stage/merge.yaml
GromNaN: Fixed.

tests/Builder/Stage/MergeStageTest.php
GromNaN: The `StringFieldPath` is not accepted where `string` is expected.
image

But phpdoc types are not expressive enough to restrict an array of FieldPath.

image

Also @alcaeus if we want to be able to alias field names, we will need objects similar to FieldPath instead of only raw strings here.

[PHPLIB-1382](https://jira.mongodb.org/browse/PHPLIB-1382)

generator/config/stage/merge.yaml
GromNaN: We could add an interface to stages that are accepted by the merge pipeline. But then we would need a `MergePipeline` class.

[PHPLIB-1383](https://jira.mongodb.org/browse/PHPLIB-1383)

tests/Builder/Stage/FacetStageTest.php
GromNaN: [PHPLIB-1384](https://jira.mongodb.org/browse/PHPLIB-1384)

tests/Builder/Stage/CurrentOpStageTest.php
GromNaN: [PHPLIB-1384](https://jira.mongodb.org/browse/PHPLIB-1384)

tests/Builder/Stage/DocumentsStageTest.php
GromNaN: To be worked in [PHPLIB-1382](https://jira.mongodb.org/browse/PHPLIB-1382)

generator/config/stage/facet.yaml
GromNaN: Fixed.

tests/Builder/Stage/GeoNearStageTest.php
GromNaN: Exactly, duplicating the example to use GeoJSON object instead of POPO.

tests/Builder/Stage/LookupStageTest.php
GromNaN: See [PHPLIB-1367](https://jira.mongodb.org/browse/PHPLIB-1367)

tests/Builder/Stage/MergeStageTest.php
alcaeus: > if we want to be able to alias field names, we will need objects similar to FieldPath instead of only raw strings here.

Yes, I was thinking the same when I saw the code. We can add that when we look at integrating the builder into ODM itself.

generator/config/stage/out.yaml
jmikola: In another [comment](https://github.com/mongodb/mongo-php-builder/pull/56/files#r1467468501), @alcaeus asked about creating custom encoder logic. Is that feasible with the current implementation, or is it more complicated because the entire OutStage file is generated (and thus would require addressing this through the generator code itself).

IIRC, the original intention of this project was to have an ongoing dependency on the generator, so I understand if there's presently no way to maintain stages and operators through PHP along (in an alternate reality where the generator was only used to create the initial implementation).

Asking the above more for my own curiosity.

I understand the current change to use `encode: single` here is the most flexible and probably sets you up to later use `string|OutCollection` for the parameter.

generator/config/stage/out.yaml
jmikola: I've no idea what this was referring to, as [$out](https://www.mongodb.com/docs/manual/reference/operator/aggregation/out/) only talks about not being able to output to a time-series collection. I don't see any reference to it as an option.

tests/Builder/Stage/ProjectStageTest.php
```diff
@@ -0,0 +1,164 @@
...
+        $this->assertSamePipeline(Pipelines::ProjectConditionallyExcludeFields, $pipeline);
+    }
+
+    public function testExcludeFieldsFromEmbeddedDocuments(): void
+    {
+        $pipeline = new Pipeline(
+            // Both stages are equivalents
+            Stage::project(
+                ...['author.first' => 0],
+                ...['lastModified' => 0],
```
jmikola, line 50: How are you using the splat operator twice, once for each positional argument? Wouldn't the correct syntax be: `...['author.first' => 0, 'lastModified' => 0]`?

generator/config/stage/set.yaml
```diff
@@ -13,3 +13,61 @@ arguments:
...
         variadic: object
+tests:
+    -
+        name: 'Using Two $set Stages'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/set/#using-two--set-stages'
+        pipeline:
+            -
+                $set:
+                    totalHomework:
+                        # The $sum expression is always build as an array, even if the value is an array field name
```
jmikola, line 24:
```suggestion
                        # The $sum expression is always built as an array, even if the value is an array field name
```

Alternatively: "The $sum expression always builds as..."

jmikola: Does the type of the referenced field factor in here? I assume `['$homework']` and `$homework` would still be computed the same and follows the [$sum: Array Operand](https://www.mongodb.com/docs/manual/reference/operator/aggregation/sum/#array-operand) rules. In other words, the array wrapping has no effect other than visually being the long form API.

generator/config/stage/set.yaml
```diff
@@ -13,3 +13,61 @@ arguments:
...
+                            -
+                                - 7
+    -
+        name: 'Creating a New Field with Existing Fields'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/set/#creating-a-new-field-with-existing-fields'
+        pipeline:
+            -
+                $set:
+                    quizAverage:
+                        # $avg: '$quiz'
```
jmikola, line 72: Add comment about `$avg` always using the long form syntax (here and elsewhere as needed).

generator/config/stage/setWindowFields.yaml
jmikola: Should this be ticketed under [PHPLIB-1381](https://jira.mongodb.org/browse/PHPLIB-1381)?

generator/config/stage/setWindowFields.yaml
```diff
@@ -28,3 +28,117 @@ arguments:
...
+    -
+        name: 'Use Documents Window to Obtain Cumulative Quantity for Each Year'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/setWindowFields/#use-documents-window-to-obtain-cumulative-quantity-for-each-year'
+        pipeline:
+            -
+                $setWindowFields:
+                    partitionBy:
+                        # $year: '$orderDate'
```
jmikola, line 53: Comment for long form explanation (here and elsewhere).

generator/config/stage/unionWith.yaml
```diff
@@ -22,3 +22,62 @@ arguments:
...
+                $sort:
+                    _id: 1
+                    store: 1
+                    item: 1
+    -
+        name: 'Report 2 Aggregated Sales by Items'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/unionWith/#report-2--aggregated-sales-by-items'
+        pipeline:
+            -
+                # $unionWith: 'sales_2018'
```
jmikola, line 65: Comments about long form syntax here.

generator/config/stage/unionWith.yaml
```diff
@@ -22,3 +22,62 @@ arguments:
...
+                    coll: 'sales_2020'
+            -
+                $group:
+                    _id: '$item'
+                    total:
+                        $sum: '$quantity'
+            -
+                $sort:
+                    total: -1
+
```
jmikola:
```suggestion
```

generator/config/stage/unset.yaml
```diff
@@ -13,3 +13,29 @@ arguments:
...
+                    - 'copies'
+    -
+        name: 'Remove Embedded Fields'
+        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/unset/#remove-embedded-fields'
+        pipeline:
+            -
+                $unset:
+                    - 'isbn'
+                    - 'author.first'
+                    - 'copies.warehouse'
```
jmikola, line 42: Same thought as https://github.com/mongodb/mongo-php-builder/pull/56#discussion_r1470137213 (re: `fieldPath()` usage), which you noted would be addressed by [PHPLIB-1382](https://jira.mongodb.org/browse/PHPLIB-1382).

generator/config/stage/merge.yaml
jmikola: Thansk for creating the ticket. I think we discussed during triage earlier, but we can decide how to handle this at a later point.

generator/config/stage/merge.yaml
jmikola: Should this be ticketed under [PHPLIB-1381](https://jira.mongodb.org/browse/PHPLIB-1381)?

jmikola (commented):
Finishing with the second half of my review. Aside from individual, open comment threads, OutStage seems to be an outstanding thing to address.

tests/Builder/Stage/ProjectStageTest.php
GromNaN: Both work. I used this syntax to keep 1 field per line.

generator/config/stage/setWindowFields.yaml
GromNaN: I already have this list of enums in [`expression.php`](https://github.com/mongodb/mongo-php-builder/blob/c0daa64a75590465d6c052ed78b60b31814bff72/generator/config/expressions.php#L121-L162). I don't feel the need to create a ticket for each.

generator/config/stage/out.yaml
GromNaN: Found here: [Use the timeseries option with the `$out` aggregation stage](https://www.mongodb.com/docs/manual/core/timeseries/timeseries-migrate-data-into-timeseries-collection/#use-the-timeseries-option-with-the--out-aggregation-stage)

https://jira.mongodb.org/browse/PHPLIB-1386

generator/config/stage/setWindowFields.yaml
jmikola: Noted. I missed those `@todo` comments.

Feel free to resolve any threads where I was asking about tickets, then -- unless there's some other open question.

generator/config/stage/out.yaml
jmikola: Thanks. This looks like a docs omission, so I've opened [DOCSP-36120](https://jira.mongodb.org/browse/DOCSP-36120). I'll defer to you if you'd like to re-introduce the `timeseries` option or defer it for a later time.

> I removed this argument, since it needs to be inside the coll object.

Looking at the example you shared above, `timeseries` is a sibling of `coll`, not a child. So I see no reason it cannot be modeled in the YAML definition.

tests/Builder/Stage/MergeStageTest.php
jmikola: @GromNaN feel free to resolve this.

tests/Builder/Stage/ProjectStageTest.php
jmikola: TIL. There is seemingly no mention of this within PHP docs (just read through [Function Arguments](https://www.php.net/manual/en/functions.arguments.php)). Tested it myself: https://3v4l.org/sJ5KB#v8.3.2

jmikola (approved):

PR #57, opened 2024-01-26
Allow to merge an array of stages into a pipeline

GromNaN:
Coming from https://github.com/mongodb/mongo-php-builder/pull/56/files#r1466461225

I feel that a list of stages is a valid type for a pipeline. We already allowed merging `Pipeline` objects, we can make things more efficient by avoiding the intermediate Pipeline object when we already have an array.

### DIFF
--- a/src/Builder/Pipeline.php
+++ b/src/Builder/Pipeline.php
@@ -8,10 +8,10 @@ use ArrayIterator;
 use IteratorAggregate;
 use MongoDB\Builder\Type\StageInterface;
 use MongoDB\Exception\InvalidArgumentException;
-use Traversable;
 
 use function array_is_list;
 use function array_merge;
+use function is_array;
 
 /**
  * An aggregation pipeline consists of one or more stages that process documents.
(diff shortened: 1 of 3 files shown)


src/Builder/Pipeline.php
```diff
@@ -36,7 +39,9 @@ public function __construct(StageInterface|Pipeline ...$stagesOrPipelines)
         $stages = [];
 
         foreach ($stagesOrPipelines as $stageOrPipeline) {
-            if ($stageOrPipeline instanceof Pipeline) {
+            if (is_array($stageOrPipeline) && array_is_list($stageOrPipeline)) {
```
alcaeus, line 44: I noticed you don't check whether each item in the list is a `StageInterface` instance. In theory, this should be fine as `PipelineEncoder` uses `encodeIfSupported` to encode each stage. Anything the encoder can't handle will be sent to the server as is, potentially causing an execution error later. This would allow users to inject raw pipeline stages (e.g. if they want to use something not supported by their builder version) as long as they know what they are doing.

alcaeus (dismissed):

alcaeus (approved):

PR #58, opened 2024-01-29
PHPLIB-1388 Fix `$search` and `$searchMeta` operators

GromNaN:
Fix [PHPLIB-1388](https://jira.mongodb.org/browse/PHPLIB-1388)

The syntax of this operator is scary.
https://www.mongodb.com/docs/atlas/atlas-search/query-syntax/


alcaeus:
Yep, these operators are quite complex, I discovered that when working on them for the Doctrine ODM. For now, I've added tests (as required by PHPLIB-1380) in #61, using `object(...)` to define the operators manually. I've also created PHPLIB-1388 to track creating builders for individual operators and adding tests for all of them.

PR #60, opened 2024-02-02
PHPLIB-1384: Use default value in Query::exists()

alcaeus:
PHPLIB-1384

I decided to add another test to explicitly test using `exists` with a different value.

### DIFF
--- a/generator/config/query/exists.yaml
+++ b/generator/config/query/exists.yaml
@@ -11,6 +11,7 @@ arguments:
         name: exists
         type:
             - bool
+        default: true
 tests:
     -
         name: 'Exists and Not Equal To'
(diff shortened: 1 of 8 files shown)


GromNaN (approved):
I'm fine with keeping the explicit `Query::exists(true)` in some tests, but we should have static analysis warnings for setting the default argument value.

https://github.com/mongodb/mongo-php-builder/blob/214baf1c9fb34656029902c2f4291da0c71ee703/tests/Builder/Query/AndOperatorTest.php#L26

https://github.com/mongodb/mongo-php-builder/blob/214baf1c9fb34656029902c2f4291da0c71ee703/tests/Builder/Stage/FacetStageTest.php#L43

https://github.com/mongodb/mongo-php-builder/blob/214baf1c9fb34656029902c2f4291da0c71ee703/tests/Builder/Stage/CurrentOpStageTest.php#L26

GromNaN (approved):

PR #62, opened 2024-02-02
PHPLIB-1273: Improve $switch syntax

alcaeus:
PHPLIB-1273

Instead of using `object`, there is now a dedicated `case` operator. In order to accommodate the different encoding for `case`, I added a `flat_object` encoding mechanism that doesn't wrap the parameter object in another object containing the operator name. Alternatively, we could create a separate encoder for `case`.

### DIFF
--- a/generator/config/expression/case.yaml
+++ b/generator/config/expression/case.yaml
@@ -2,7 +2,7 @@
 name: $case
 link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/switch/'
 type:
-    - resolvesToAny
+    - switchBranch
 encode: flat_object
 description: |
     Represents a single case in a $switch expression
--- a/generator/config/expression/case.yaml
+++ b/generator/config/expression/case.yaml
@@ -1,0 +1,21 @@
+# $schema: ../schema.json
+name: $case
+link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/switch/'
+type:
+    - resolvesToAny
+encode: flat_object
+description: |
+    Represents a single case in a $switch expression
+arguments:
+    -
+        name: case
+        type:
+            - resolvesToBool
+        description: |
+            Can be any valid expression that resolves to a boolean. If the result is not a boolean, it is coerced to a boolean value. More information about how MongoDB evaluates expressions as either true or false can be found here.
+    -
+        name: then
+        type:
+            - expression
+        description: |
+            Can be any valid expression.
--- a/generator/config/schema.json
+++ b/generator/config/schema.json
@@ -58,7 +58,7 @@
                     "type": "string",
                     "enum": [
                         "array",
-                        "object",
+                        "flat_object",
                         "dollar_object",
                         "single",
                         "group"
--- a/generator/config/schema.json
+++ b/generator/config/schema.json
@@ -29,6 +29,7 @@
                             "filter",
                             "window",
                             "geometry",
+                            "switchBranch",
                             "resolvesToAny",
                             "resolvesToNumber",
                             "resolvesToDouble",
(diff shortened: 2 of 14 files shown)


generator/config/expression/case.yaml
```diff
@@ -0,0 +1,21 @@
+# $schema: ../schema.json
+name: $case
+link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/switch/'
```
GromNaN, line 3: I see the interest of creating this "fake" operator for code generation. It should be consistent with [`OutputWindow`](https://github.com/mongodb/mongo-php-builder/blob/592b0a0a1c60b3e412c1cfa91da8262c7a83fd72/src/Builder/Accumulator.php#L30-L37) this is the same need.

Could you create `accumulator/outputWindow.yaml` also?

generator/config/schema.json
```diff
@@ -58,7 +58,7 @@
                     "type": "string",
                     "enum": [
                         "array",
-                        "object",
+                        "flat_object",
```
GromNaN, line 63: `object` is still used.
```suggestion
                        "object",
                        "flat_object",
```

generator/config/expression/case.yaml
```diff
@@ -0,0 +1,21 @@
...
+type:
+    - resolvesToAny
```
GromNaN: It should resolve to a new type "switchCase", as it can't be used anywhere an expression is expected.

src/Builder/Encoder/OperatorEncoder.php
```diff
@@ -95,7 +98,7 @@ private function encodeAsGroup(GroupStage $value): stdClass
         return $this->wrap($value, $result);
     }
 
-    private function encodeAsObject(OperatorInterface $value): stdClass
+    private function encodeAsObject(OperatorInterface $value, bool $flat = false): stdClass
```
GromNaN: The `$flat` behavior is already hold by `$value`.

I would modify the `wrap` method to skip if `$operator::getOperator()` returns `null`.

alcaeus: We can't have `getOperator()` return `null`: the only way to do so would be to set the operator name to `null` in the config, which then causes issues as the name of the generated class depends on that as well.

generator/config/expression/case.yaml
alcaeus: I'll do so in a separate PR: the current `OutputWindow` type contains additional validation logic that we can't represent through the config.

generator/config/schema.json
alcaeus: Whoops. Done.

generator/config/expression/case.yaml
alcaeus: Added a new `SwitchBranch` interface for this and defined it as a type in the generator.

src/Builder/Encoder/OperatorEncoder.php
GromNaN: Oh right. What do you think of checking `$value::ENCODE` in `wrap`?

GromNaN (commented):
Switch stage is also in `$convert`, `$isNumber` and `$toBool` operator tests.

alcaeus:
> Switch stage is also in `$convert`, `$isNumber` and `$toBool` operator tests.

Updated

src/Builder/Encoder/OperatorEncoder.php
alcaeus: I added that. It doesn't make much difference, but I guess it's a little cleaner. We might end up splitting up `OperatorEncoder` further, but for now this works.

GromNaN (approved):

PR #63, opened 2024-02-02
PHPLIB-1381 PHPLIB-1269 Add enum for Sort and TimeUnit

GromNaN:
Fix [PHPLIB-1381](https://jira.mongodb.org/browse/PHPLIB-1381) and [PHPLIB-1269](https://jira.mongodb.org/browse/PHPLIB-1269)

I decided to not rely on `BackedEnum`, but use a specific `Dictionary` interface, with a method `getValue()` so that the value can be different than only `string` or only `int`. This is the case for `Sort::TextScore`.

### DIFF
--- a/src/Builder/Type/Sort.php
+++ b/src/Builder/Type/Sort.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Builder\Type;
+
+enum Sort implements Dictionnary
+{
+    case Asc;
+    case Desc;
+    case TextScore;
+    case SearchScoreAsc;
+    case SearchScoreDesc;
+
+    public function getValue(): int|array
+    {
+        return match ($this) {
+            self::Asc => 1,
+            self::Desc => -1,
+            self::TextScore => ['$meta' => 'textScore'],
+            self::SearchScoreAsc => ['$meta' => 'searchScore', 'order' => 1],
+            self::SearchScoreDesc => ['$meta' => 'searchScore', 'order' => -1],
+        };
+    }
+}
--- a/src/Builder/Type/Sort.php
+++ b/src/Builder/Type/Sort.php
@@ -14,8 +14,6 @@ enum Sort implements DictionaryInterface
     case Asc;
     case Desc;
     case TextScore;
-    case SearchScoreAsc;
-    case SearchScoreDesc;
 
     public function getValue(): int|array
     {
@@ -23,8 +21,6 @@ enum Sort implements DictionaryInterface
             self::Asc => 1,
             self::Desc => -1,
             self::TextScore => ['$meta' => 'textScore'],
-            self::SearchScoreAsc => ['$meta' => 'searchScore', 'order' => 1],
-            self::SearchScoreDesc => ['$meta' => 'searchScore', 'order' => -1],
         };
     }
 }
--- a/src/Builder/Type/Sort.php
+++ b/src/Builder/Type/Sort.php
@@ -4,7 +4,12 @@ declare(strict_types=1);
 
 namespace MongoDB\Builder\Type;
 
-enum Sort implements Dictionnary
+/**
+ * Sort order can be used with $sort stage and sortBy properties
+ *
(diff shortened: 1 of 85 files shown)


src/Builder/Type/Sort.php
```diff
@@ -0,0 +1,25 @@
...
+    case SearchScoreDesc;
+
+    public function getValue(): int|array
+    {
+        return match ($this) {
+            self::Asc => 1,
+            self::Desc => -1,
+            self::TextScore => ['$meta' => 'textScore'],
+            self::SearchScoreAsc => ['$meta' => 'searchScore', 'order' => 1],
+            self::SearchScoreDesc => ['$meta' => 'searchScore', 'order' => -1],
```
GromNaN: See https://www.mongodb.com/docs/atlas/atlas-search/sort/#sort-by-score-and-date-or-numeric-field

src/Builder/Type/Sort.php
```diff
@@ -0,0 +1,30 @@
...
```
alcaeus: Does it make sense to include all of these in a single `Sort` enum? IIRC `searchScore` is only available in `$search` but not `$sort`, while `textScore` is only available in `$sort` but not `$search`.

src/Builder/Type/TimeUnit.php
```diff
@@ -0,0 +1,26 @@
...
+{
+    case Year = 'year';
+    case Quarter = 'quarter';
+    case Week = 'week';
+    case Month = 'month';
+    case Day = 'day';
+    case Hour = 'hour';
+    case Minute = 'minute';
+    case Second = 'second';
+    case Millisecond = 'millisecond';
```
alcaeus, line 20: Nit: I'd argue the correct capitalisation would be `MilliSecond`. This has the added advantage of PhpStorm suggesting it when typing `TimeUnit::ms` in the autocompletion box.

tests/Builder/Stage/SortStageTest.php
```diff
@@ -20,10 +19,8 @@ public function testAscendingDescendingSort(): void
     {
         $pipeline = new Pipeline(
             Stage::sort(
-                object(
-                    age: -1,
-                    posts: 1,
-                ),
+                age: Sort::Desc,
+                posts: Sort::Asc,
```
alcaeus, line 23: 😍

alcaeus (commented):
Minor feedback regarding the enums, but in general LGTM!

src/Builder/Type/Sort.php
GromNaN: Ok, removed.

src/Builder/Type/TimeUnit.php
GromNaN: Done. But I don't get any completion for `TimeUnit::ms` in PHPStorm. Whereas I have completion for both `TimeUnit::Ms` and `TimeUnit::MS` with or without this change. Only the upper M matters.

alcaeus: Thanks for confirming - I had hoped a miniscule "m" would already trigger the correct selection. Thinking about this some more, it should probably be "Millisecond" after all, as "MilliSecond" would imply "milli second" as the correct unit, which is definitely wrong. Sorry for the back and forth here. 😨

src/Builder/Type/Sort.php
alcaeus: Do we want to include `searchScore` in a different enum or leave it to the task of adding full operator support to `$search`? If so, let's make a note in the respective ticket to remind ourselves.

GromNaN: Reverted.

src/Builder/Type/TimeUnit.php
GromNaN: Reverted.

src/Builder/Type/Sort.php
GromNaN: Let's keep the new enum for later in https://github.com/mongodb/mongo-php-builder/pull/58

alcaeus (approved):
LGTM!

PR #64, opened 2024-02-02
PHPLIB-1367 Add enum for System variables

GromNaN:
Fix [PHPLIB-1367](https://jira.mongodb.org/browse/PHPLIB-1367)

Includes https://github.com/mongodb/mongo-php-builder/pull/63 (to be merged before)

I originally created factories for system variables, but we can also use enums. I'm not sure what's the best approach. They are almost the same for DX


alcaeus (approved):

src/Builder/Variable.php
```diff
@@ -1,158 +0,0 @@
...
-        return new Expression\Variable('ROOT');
-    }
-
-    /**
-     * References the start of the field path being processed in the aggregation pipeline stage.
-     * Unless documented otherwise, all stages start with CURRENT the same as ROOT.
-     * CURRENT is modifiable. However, since $<field> is equivalent to $$CURRENT.<field>, rebinding
-     * CURRENT changes the meaning of $ accesses.
-     */
-    public static function current(string $fieldPath = ''): ResolvesToAny
```
GromNaN, line 61: This helper was more explicit than the new `Variable::Current->dot(...)` function in introduced to the enum.

tests/Builder/Expression/UnsetFieldOperatorTest.php
```diff
@@ -38,7 +39,7 @@ public function testRemoveFieldsThatContainPeriods(): void
             Stage::replaceWith(
                 Expression::unsetField(
                     field: 'price.usd',
-                    input: Expression::variable('ROOT'),
+                    input: Variable::Root,
```
GromNaN, line 42: This could have been `Variable::root()`, which is as good as the enum `Variable::Root`.

GromNaN (commented):
I'm still not convinced this approach is better than the factory methods.

GromNaN:
Closing for now. Lets revisit later. Ticket still open.

PR #65, opened 2024-02-26
Add fluent builder for Laravel

GromNaN:
Related to PHPORM-103

Create a fluent pipeline builder that can be used like this:

```php
        $pipeline = (new FluentFactory())
            ->match(x: Query::eq(1))
            ->project(_id: false, x: true)
            ->sort(x: Sort::Asc)
            ->getPipeline();
```

### DIFF
--- a/src/Builder/Type/ExpressionInterface.php
+++ b/src/Builder/Type/ExpressionInterface.php
@@ -8,7 +8,7 @@ namespace MongoDB\Builder\Type;
  * Expressions can include field paths, literals, system variables, expression objects, and expression operators.
  * Expressions can be nested.
  *
- * @see https://www.mongodb.com/docs/manual/meta/aggregation-quick-reference/#expressions
+ * @see https://www.mongodb.com/docs/manual/reference/operator/aggregation/
  */
 interface ExpressionInterface
 {
--- a/src/Builder/Type/StageInterface.php
+++ b/src/Builder/Type/StageInterface.php
@@ -7,7 +7,7 @@ namespace MongoDB\Builder\Type;
 /**
  * Aggregation Pipeline Stages
  *
- * @see https://www.mongodb.com/docs/manual/meta/aggregation-quick-reference/#stages
+ * @see https://www.mongodb.com/docs/manual/reference/operator/aggregation-pipeline/
  */
 interface StageInterface
 {


generator/src/FluentStageFactoryGenerator.php
```diff
@@ -0,0 +1,158 @@
...
+ * The method definition is based on the manually edited static class
+ * that imports the stage factory trait.
+ */
+class FluentStageFactoryGenerator extends OperatorGenerator
+{
+    /**
+     * All public of this class are duplicated as instance methods of the
+     * fluent factory class.
+     */
+    private const FACTORY_CLASS = Stage::class;
```
GromNaN: In order to get the customized signature of some factory methods, we read the signature from `MongoDB\Builder\Stage` static methods.

generator/src/FluentStageFactoryGenerator.php
```diff
@@ -0,0 +1,158 @@
...
+        $namespace->addUse(PackedArray::class);
+        $namespace->addUse(Serializable::class);
+        $namespace->addUse(Timestamp::class);
+        $namespace->addUse(Type::class);
+        $namespace->addUse(ArrayFieldPath::class);
+        $namespace->addUse(FieldPath::class);
+        $namespace->addUse(ResolvesToArray::class);
+        $namespace->addUse(ResolvesToObject::class);
+        $namespace->addUse(Pipeline::class);
+        $namespace->addUse(AccumulatorInterface::class);
```
GromNaN: This long list should be read from the source code.

generator/src/FluentStageFactoryGenerator.php
```diff
@@ -0,0 +1,163 @@
...
+        $method->setReturnType('static');
+        $method->setBody(sprintf(
+            <<<'PHP'
+            $this->pipeline[] = %s::%s(%s);
+
+            return $this;
+            PHP,
+            'Stage',
+            $factoryMethod->getName(),
+            implode(',', $args),
```
GromNaN: Space missing.
```suggestion
            implode(', ', $args),
```

generator/src/FluentStageFactoryGenerator.php
```diff
@@ -0,0 +1,163 @@
...
+            }
+
+            try {
+                $this->addMethod($method, $namespace, $class);
+            } catch (Throwable $e) {
+                throw new RuntimeException(sprintf('Failed to generate class for operator "%s"', $operator->name), 0, $e);
+            }
+        }
+
+        $staticFactory = TraitType::from(Stage\FactoryTrait::class);
```
alcaeus: I noticed that the factory class was extracted to a private constant, but this trait was not. Should this be extracted?

generator/src/FluentStageFactoryGenerator.php
alcaeus: Does the generated code include fully qualified names without these use statements, or would the code break? IMO, in generated code it's fine to not rely on `use` statements, alternatively we could always generate the code with fully qualified names and then let phpcbf fix whatever it can (without actually checking the code against our styleguide so that manual fixes need not to be applied).

generator/src/FluentStageFactoryGenerator.php
```diff
@@ -0,0 +1,163 @@
...
+            ->setBody(<<<'PHP'
+                return new Pipeline(...$this->pipeline);
+                PHP);
+
+        $staticFactory = ClassType::from(self::FACTORY_CLASS);
+        assert($staticFactory instanceof ClassType);
+        foreach ($staticFactory->getMethods() as $method) {
+            if (! $method->isPublic()) {
+                continue;
+            }
```
alcaeus: Moving this check into `addMethod` would allow you to simplify the entire block:
```php
array_map(
    fn ($method) => $this->addMethod($method, $namespace, $class),
    $staticFactory->getMethods(),
);
```

tests/Builder/FluentPipelineFactoryTest.php
```diff
@@ -0,0 +1,26 @@
...
+    public function testFluentPipelineFactory(): void
+    {
+        $pipeline = (new FluentFactory())
+            ->match(x: Query::eq(1))
+            ->project(_id: false, x: true)
+            ->sort(x: Sort::Asc)
+            ->getPipeline();
+
+        $this->assertInstanceof(Pipeline::class, $pipeline);
+        $this->assertCount(3, $pipeline->getIterator());
```
alcaeus: Can you also add an assertion to ensure that the correct stages are added? As is we're only checking that three stages are added, but not what is actually being added.
A possible alternative would be to duplicate the existing stage tests and creating them with the fluent builder instead of the static one, e.g.:
```php
public function testEqualityMatch(): void
{
    $pipeline = (new FluentFactory())
        ->match(author: 'dave'),
        ->getPipeline();

    $this->assertSamePipeline(Pipelines::MatchEqualityMatch, $pipeline);
}
```

alcaeus (changes requested):

generator/src/FluentStageFactoryGenerator.php
GromNaN: This breaks phpdoc.

generator/src/FluentStageFactoryGenerator.php
GromNaN: Good point. I move the `isPublic` check but keep the `foreach`.

generator/src/FluentStageFactoryGenerator.php
GromNaN: Changed to use reflection to get the trait.

tests/Builder/FluentPipelineFactoryTest.php
GromNaN: I don't think replicating all tests is interesting. This single test can cover all specific cases.

GromNaN:
I messed with my branches. PR re-opened: https://github.com/mongodb/mongo-php-builder/pull/67

PR #67, opened 2024-03-12
Add fluent builder for Laravel 

GromNaN:
Related to PHPORM-103

Reopen #65

Create a fluent pipeline builder that can be used like this:

```php
        $pipeline = (new class { use FluentFactoryTrait; })
            ->match(x: Query::eq(1))
            ->project(_id: false, x: true)
            ->sort(x: Sort::Asc)
            ->getPipeline();
```

### DIFF
--- a/generator/src/FluentStageFactoryGenerator.php
+++ b/generator/src/FluentStageFactoryGenerator.php
@@ -24,7 +24,7 @@ use function implode;
 use function sprintf;
 
 /**
- * Generates a fluent factory class for aggregation pipeline stages.
+ * Generates a fluent factory trait for aggregation pipeline stages.
  * The method definition is based on all the public static methods
  * of the Stage class.
  */
@@ -34,23 +34,23 @@ class FluentStageFactoryGenerator extends OperatorGenerator
 
     public function generate(GeneratorDefinition $definition): void
     {
-        $this->writeFile($this->createFluentFactoryClass($definition));
+        $this->writeFile($this->createFluentFactoryTrait($definition));
     }
 
-    private function createFluentFactoryClass(GeneratorDefinition $definition): PhpNamespace
+    private function createFluentFactoryTrait(GeneratorDefinition $definition): PhpNamespace
     {
         $namespace = new PhpNamespace($definition->namespace);
-        $class = $namespace->addClass('FluentFactory');
+        $trait = $namespace->addTrait('FluentFactoryTrait');
 
         $namespace->addUse(self::FACTORY_CLASS);
         $namespace->addUse(StageInterface::class);
         $namespace->addUse(Pipeline::class);
 
-        $class->addProperty('pipeline')
+        $trait->addProperty('pipeline')
             ->setType('array')
             ->setComment('@var list<StageInterface>')
             ->setValue([]);
-        $class->addMethod('getPipeline')
+        $trait->addMethod('getPipeline')
             ->setReturnType(Pipeline::class)
             ->setBody(<<<'PHP'
                 return new Pipeline(...$this->pipeline);
@@ -62,23 +62,23 @@ class FluentStageFactoryGenerator extends OperatorGenerator
 
         // Import the methods customized in the factory class
         foreach ($staticFactory->getMethods() as $method) {
-            $this->addMethod($method, $class);
+            $this->addMethod($method, $trait);
         }
 
         // Import the other methods provided by the generated trait
-        foreach ($staticFactory->getTraits() as $trait) {
-            $this->addUsesFrom($trait->getName(), $namespace);
-            $staticFactory = TraitType::from($trait->getName());
+        foreach ($staticFactory->getTraits() as $usedTrait) {
+            $this->addUsesFrom($usedTrait->getName(), $namespace);
+            $staticFactory = TraitType::from($usedTrait->getName());
             assert($staticFactory instanceof TraitType);
             foreach ($staticFactory->getMethods() as $method) {
-                $this->addMethod($method, $class);
+                $this->addMethod($method, $trait);
             }
         }
 
         return $namespace;
     }
 
-    private function addMethod(Method $factoryMethod, ClassType $class): void
+    private function addMethod(Method $factoryMethod, TraitType $trait): void
     {
         // Non-public methods are not part of the API
         if (! $factoryMethod->isPublic()) {
@@ -87,11 +87,11 @@ class FluentStageFactoryGenerator extends OperatorGenerator
 
         // Some methods can be overridden in the class, so we skip them
         // when importing the methods provided by the trait.
-        if ($class->hasMethod($factoryMethod->getName())) {
+        if ($trait->hasMethod($factoryMethod->getName())) {
             return;
         }
 
-        $method = $class->addMethod($factoryMethod->getName());
+        $method = $trait->addMethod($factoryMethod->getName());
 
         $method->setComment($factoryMethod->getComment());
         $method->setParameters($factoryMethod->getParameters());
@@ -119,9 +119,9 @@ class FluentStageFactoryGenerator extends OperatorGenerator
         ));
     }
 
-    private static function addUsesFrom(string $class, PhpNamespace $namespace): void
+    private static function addUsesFrom(string $classLike, PhpNamespace $namespace): void
     {
-        $file = PhpFile::fromCode(file_get_contents((new ReflectionClass($class))->getFileName()));
+        $file = PhpFile::fromCode(file_get_contents((new ReflectionClass($classLike))->getFileName()));
 
         foreach ($file->getNamespaces() as $ns) {
             foreach ($ns->getUses() as $use) {
--- a/generator/src/FluentStageFactoryGenerator.php
+++ b/generator/src/FluentStageFactoryGenerator.php
@@ -4,39 +4,22 @@ declare(strict_types=1);
 
 namespace MongoDB\CodeGenerator;
 
-use MongoDB\BSON\Decimal128;
-use MongoDB\BSON\Document;
-use MongoDB\BSON\Int64;
-use MongoDB\BSON\PackedArray;
-use MongoDB\BSON\Serializable;
-use MongoDB\BSON\Timestamp;
-use MongoDB\BSON\Type;
-use MongoDB\Builder\Expression\ArrayFieldPath;
-use MongoDB\Builder\Expression\FieldPath;
-use MongoDB\Builder\Expression\ResolvesToArray;
-use MongoDB\Builder\Expression\ResolvesToObject;
 use MongoDB\Builder\Pipeline;
 use MongoDB\Builder\Stage;
-use MongoDB\Builder\Type\AccumulatorInterface;
-use MongoDB\Builder\Type\ExpressionInterface;
-use MongoDB\Builder\Type\FieldQueryInterface;
-use MongoDB\Builder\Type\Optional;
-use MongoDB\Builder\Type\QueryInterface;
-use MongoDB\Builder\Type\Sort;
 use MongoDB\Builder\Type\StageInterface;
 use MongoDB\CodeGenerator\Definition\GeneratorDefinition;
-use MongoDB\Model\BSONArray;
 use Nette\PhpGenerator\ClassType;
 use Nette\PhpGenerator\Method;
 use Nette\PhpGenerator\Parameter;
+use Nette\PhpGenerator\PhpFile;
 use Nette\PhpGenerator\PhpNamespace;
 use Nette\PhpGenerator\TraitType;
 use ReflectionClass;
-use stdClass;
 
 use function array_key_last;
 use function array_map;
 use function assert;
+use function file_get_contents;
 use function implode;
 use function sprintf;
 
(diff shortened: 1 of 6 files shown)


generator/src/FluentStageFactoryGenerator.php
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ * Generates a fluent factory class for aggregation pipeline stages.
+ * The method definition is based on the manually edited static class
+ * that imports the stage factory trait.
+ */
+class FluentStageFactoryGenerator extends OperatorGenerator
+{
+    /**
+     * All public of this class are duplicated as instance methods of the
```
jmikola: I think you accidentally a word here. Should it be "public _methods_"?

generator/src/FluentStageFactoryGenerator.php
```diff
@@ -0,0 +1,163 @@
...
+     * fluent factory class.
+     */
+    private const FACTORY_CLASS = Stage::class;
```
jmikola, line 33: Is there a reason for making this a constant? I see it's only referenced twice in `createFluentFactoryClass()`.

generator/src/FluentStageFactoryGenerator.php
```diff
@@ -0,0 +1,163 @@
...
+            }
+        }
+
+        return $namespace;
+    }
+
+    private function addMethod(Method $factoryMethod, PhpNamespace $namespace, ClassType $class): void
+    {
+        if ($class->hasMethod($factoryMethod->getName())) {
+            return;
```
jmikola, line 85: Would it make sense to throw here? What is the legitimate case where the method already exists and you'd want to NOP?

generator/src/FluentStageFactoryGenerator.php
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        $method->setReturnType('static');
+        $method->setBody(sprintf(
+            <<<'PHP'
+            $this->pipeline[] = %s::%s(%s);
+
+            return $this;
+            PHP,
+            'Stage',
```
jmikola: I assume this is meant to correspond to `FACTORY_CLASS`. If so, would it make sense to derive it from that instead of hard-coding its short name? That could probably be done once with:

```php
static $factoryClass = (new ReflectionClass(self::FACTORY_CLASS))->getShortName();
```

generator/src/FluentStageFactoryGenerator.php
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        $method = $class->addMethod($factoryMethod->getName());
+
+        $method->setComment($factoryMethod->getComment());
+        $method->setParameters($factoryMethod->getParameters());
+
+        $args = array_map(
+            fn ($param) => '$' . $param->getName(),
+            $factoryMethod->getParameters(),
```
jmikola, line 101: Looking at [`FunctionLike::getParameters()`](https://github.com/nette/php-generator/blob/v4.1.4/src/PhpGenerator/Traits/FunctionLike.php#L81), I see this returns an array of `Nette\PhpGenerator\Parameter` objects. Should you use that type-hint in the closure above?

Also, would a `string` return be helpful for static analysis? That's less important to me, as it's clear to the reader that this returns a string via concatenation.

generator/src/FluentStageFactoryGenerator.php
```diff
@@ -0,0 +1,163 @@
...
+use MongoDB\BSON\PackedArray;
+use MongoDB\BSON\Serializable;
+use MongoDB\BSON\Timestamp;
+use MongoDB\BSON\Type;
+use MongoDB\Builder\Expression\ArrayFieldPath;
+use MongoDB\Builder\Expression\FieldPath;
+use MongoDB\Builder\Expression\ResolvesToArray;
+use MongoDB\Builder\Expression\ResolvesToObject;
+use MongoDB\Builder\Pipeline;
+use MongoDB\Builder\Stage;
```
jmikola, line 8: This refers to [src/Builder/Stage.php](https://github.com/mongodb/mongo-php-builder/blob/0.1/src/Builder/Stage.php). But later on, you also have a reference to `Stage\FactoryTrait`, which seems to piggy-back this `use` statement. Is there any reason not to import that trait directly?

I forgot that [namespaces can be imported](https://www.php.net/manual/en/language.namespaces.importing.php), as it seldom comes up in my work.

generator/src/FluentStageFactoryGenerator.php
```diff
@@ -0,0 +1,163 @@
...
+use function array_key_last;
+use function array_map;
+use function assert;
+use function implode;
+use function sprintf;
+
+/**
+ * Generates a fluent factory class for aggregation pipeline stages.
+ * The method definition is based on the manually edited static class
+ * that imports the stage factory trait.
```
jmikola: I see that `createFluentFactoryClass()` proxies all public methods from the factory class _and_ `Stage\FactoryTrait`.

Can you elaborate on what the "manually edited static class" is, and where the "stage factory trait" is imported? I'm definitely missing some context here (it's been a while since I reviewed the original PRs for this project).

generator/src/FluentStageFactoryGenerator.php
```diff
@@ -0,0 +1,163 @@
...
+        $namespace->addUse(ResolvesToArray::class);
+        $namespace->addUse(ResolvesToObject::class);
+        $namespace->addUse(Pipeline::class);
+        $namespace->addUse(AccumulatorInterface::class);
+        $namespace->addUse(ExpressionInterface::class);
+        $namespace->addUse(Optional::class);
+        $namespace->addUse(QueryInterface::class);
+        $namespace->addUse(Sort::class);
+        $namespace->addUse(BSONArray::class);
+        $namespace->addUse(stdClass::class);
```
jmikola: The list of imports here seems fragile. I assume it needs to be a union of imports found throughout the factory class and trait. Would it make more sense to dynamically pull those from the file?

I suppose either way, it's possible you end up importing more than is actually needed by public function signatures. I'm not sure if the generated file would get validated by CS to catch that.

generator/src/FluentStageFactoryGenerator.php
```diff
@@ -0,0 +1,163 @@
...
+        $staticFactory = ClassType::from(self::FACTORY_CLASS);
+        assert($staticFactory instanceof ClassType);
+        foreach ($staticFactory->getMethods() as $method) {
+            if (! $method->isPublic()) {
+                continue;
+            }
+
+            try {
+                $this->addMethod($method, $namespace, $class);
+            } catch (Throwable $e) {
```
jmikola: Realistically, what kind of exceptions would you expect here? `addMethod()` currently doesn't throw, so I assume it'd have to be something from php-generator's API.

GromNaN: Removed this useless try-catch. It comes from the other generator classes and is necessary to provide context. But not necessary in this class.

generator/src/FluentStageFactoryGenerator.php
GromNaN: Comment moved to the class.

generator/src/FluentStageFactoryGenerator.php
GromNaN: I'm using a constant to highlight the class that is used as base for this generator.

generator/src/FluentStageFactoryGenerator.php
GromNaN: The method `match` is redefined in the `Stage` class. The generated version from the trait must be ignored.

generator/src/FluentStageFactoryGenerator.php
GromNaN: Good point. Fixed.

generator/src/FluentStageFactoryGenerator.php
GromNaN: Good point. Fixed.

generator/src/FluentStageFactoryGenerator.php
GromNaN: The trait is now found by inspection of the `Stage` class.

generator/src/FluentStageFactoryGenerator.php
GromNaN: Comment updated.

generator/src/FluentStageFactoryGenerator.php
GromNaN: php-generator doesn't allow to get the file from a class declaration. So I used reflection to find the file name for the class and the trait, parse this files to extract the `use` statements.

generator/src/FluentStageFactoryGenerator.php
```diff
@@ -0,0 +1,136 @@
...
+        $this->addUsesFrom(self::FACTORY_CLASS, $namespace);
+        $staticFactory = ClassType::from(self::FACTORY_CLASS);
+        assert($staticFactory instanceof ClassType);
+
+        // Import the methods customized in the factory class
+        foreach ($staticFactory->getMethods() as $method) {
+            if (! $method->isPublic()) {
+                continue;
+            }
+
```
alcaeus: This can be removed now that the check is in `addMethod`:
```suggestion
```

GromNaN: Oups.

src/Builder/Stage/FluentFactoryTrait.php
```diff
@@ -0,0 +1,792 @@
...
+use MongoDB\Builder\Type\ExpressionInterface;
+use MongoDB\Builder\Type\FieldQueryInterface;
+use MongoDB\Builder\Type\Optional;
+use MongoDB\Builder\Type\QueryInterface;
+use MongoDB\Builder\Type\Sort;
+use MongoDB\Builder\Type\StageInterface;
+use MongoDB\Model\BSONArray;
+use stdClass;
+
+trait FluentFactoryTrait
```
GromNaN, line 34: I converted the class to a trait, this allows to rename or change visibility of a method if necessary, and avoid issues that can be caused by inheritance.

generator/src/FluentStageFactoryGenerator.php
jmikola: Noted. In this case, a comment would be helpful to clarify that you're giving preference to the redefined Stage methods instead of those from the trait. And this assumes you add the Stage methods _before_ the trait (no need to repeat that, it's clear from the order of `addMethod()` calls in the generation method).

generator/src/FluentStageFactoryGenerator.php
```diff
@@ -0,0 +1,132 @@
...
+use function array_key_last;
+use function array_map;
+use function assert;
+use function file_get_contents;
+use function implode;
+use function sprintf;
+
+/**
+ * Generates a fluent factory class for aggregation pipeline stages.
```
jmikola:
```suggestion
 * Generates a fluent factory trait for aggregation pipeline stages.
```

Make a note to update other references of "class" to "trait" below. For instance, `createFluentFactoryClass()` should now be `createFluentFactoryTrait()`.

This very class might also be better named `FluentFactoryTraitGenerator()`.

src/Builder/Stage/FluentFactoryTrait.php
```diff
@@ -0,0 +1,792 @@
+<?php
+
+/**
+ * THIS FILE IS AUTO-GENERATED. ANY CHANGES WILL BE LOST!
```
jmikola, line 4: If this file is auto-generated, should it be labeled as such in `.gitattributes`?

jmikola (approved):
LGTM with the class/trait name replacements. Also note a suggestion to comment the `addMethod()` return statement.

generator/src/FluentStageFactoryGenerator.php
GromNaN: That's essentially what I said in the comment above.

generator/src/FluentStageFactoryGenerator.php
GromNaN: I want to keep "Stage" in the generator class name.

src/Builder/Stage/FluentFactoryTrait.php
GromNaN: We already have `/src/Builder/Stage/*.php linguist-generated=true`

generator/src/FluentStageFactoryGenerator.php
jmikola: Oh, my mistake. I just realized this entire thread was incorrectly applied to the early return for non-public methods.

I see the comment below for skipping overridden methods. :+1:

src/Builder/Stage/FluentFactoryTrait.php
jmikola: Noted. I'm not sure why GitHub opted not to present that explanation for hiding the file (maybe "large diffs" take precedence).

jmikola (approved):

generator/src/FluentStageFactoryGenerator.php
jmikola: Should the generated file be `FluentStageFactoryTrait.php` in that case? I'll defer to you, as I'm not sure if you have future plans for more traits.

GromNaN: The file is in the Stage namespace.

PR #68, opened 2024-03-13
Fix MongoDB namespace for `InvalidArgumentException`

GromNaN:
An extra backslash and the wrong exception class was used.

I realize that we don't have tests on exceptions.

### DIFF
--- a/generator/src/OperatorClassGenerator.php
+++ b/generator/src/OperatorClassGenerator.php
@@ -78,9 +78,10 @@ class OperatorClassGenerator extends OperatorGenerator
                 $constuctor->addComment('@param ' . $type->doc . ' ...$' . $argument->name . rtrim(' ' . $argument->description));
 
                 if ($argument->variadicMin > 0) {
+                    $namespace->addUse(InvalidArgumentException::class);
                     $constuctor->addBody(<<<PHP
                     if (\count(\${$argument->name}) < {$argument->variadicMin}) {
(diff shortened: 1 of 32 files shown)


generator/src/OperatorClassGenerator.php
```diff
@@ -78,9 +78,10 @@ public function createClass(GeneratorDefinition $definition, OperatorDefinition
                 $constuctor->addComment('@param ' . $type->doc . ' ...$' . $argument->name . rtrim(' ' . $argument->description));
 
                 if ($argument->variadicMin > 0) {
+                    $namespace->addUse(InvalidArgumentException::class);
                     $constuctor->addBody(<<<PHP
                     if (\count(\${$argument->name}) < {$argument->variadicMin}) {
-                        throw new \InvalidArgumentException(\sprintf('Expected at least %d values for \${$argument->name}, got %d.', {$argument->variadicMin}, \count(\${$argument->name})));
+                        throw new InvalidArgumentException(\sprintf('Expected at least %d values for \${$argument->name}, got %d.', {$argument->variadicMin}, \count(\${$argument->name})));
```
jmikola, line 84: Do you mean to throw a `MongoDB\Exception\InvalidArgumentException` here (based on the current import) or a generic, non-vendored exception? I suppose it doesn't really matter, but this diff dropped the leading `\` on the class name.

jmikola (approved):
Just confirm what exception you want to throw from the generator. The rest LGTM.

generator/src/OperatorClassGenerator.php
GromNaN: Yes I want to throw the MongoDB exception.

PR #71, opened 2024-03-22
Fix accepted stage type in FluentFactoryTrait

GromNaN:
Fix https://github.com/mongodb/laravel-mongodb/pull/2738#discussion_r1536068328

Get the stage type accepted by `Pipeline`.
https://github.com/mongodb/mongo-php-builder/blob/aa4f34cde35608b328f23d8fa6270b7cafbe5b18/src/Builder/Pipeline.php#L23

### DIFF
--- a/generator/src/FluentStageFactoryGenerator.php
+++ b/generator/src/FluentStageFactoryGenerator.php
@@ -15,6 +15,7 @@ use Nette\PhpGenerator\PhpFile;
 use Nette\PhpGenerator\PhpNamespace;
 use Nette\PhpGenerator\TraitType;
 use ReflectionClass;
+use stdClass;
 
 use function array_key_last;
 use function array_map;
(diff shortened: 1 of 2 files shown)


generator/src/FluentStageFactoryGenerator.php
```diff
@@ -45,10 +46,11 @@ private function createFluentFactoryTrait(GeneratorDefinition $definition): PhpN
         $namespace->addUse(self::FACTORY_CLASS);
         $namespace->addUse(StageInterface::class);
         $namespace->addUse(Pipeline::class);
+        $namespace->addUse(stdClass::class);
 
         $trait->addProperty('pipeline')
             ->setType('array')
-            ->setComment('@var list<StageInterface>')
+            ->setComment('@var list<StageInterface|array<string,mixed>|stdClass>')
```
jmikola, line 53: Why was `stdClass` added? In https://github.com/mongodb/laravel-mongodb/pull/2738/files#r1536149792, you only needed `array<string,mixed>`.

GromNaN: Because that's what is accepted by the Pipeline class.

jmikola: An existing Pipeline object is also accepted by the Pipeline constructor (for appending IIRC). I just assumed this would be as restrictive as possible based on what the trait enables users to do.

Defer to you.

jmikola (approved):

generator/src/FluentStageFactoryGenerator.php
GromNaN: The property is public. The user can append anything to the array.

